PR #3, opened 2020-07-02
List recommendations

ViktoriiaKovalova:


pkg/automation/listrecommendations_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+type SleepingService struct {
+	zones                                 []string
+	numberOfTimesListRecommendationsCalls int
+	requestsDurations                     []time.Duration
+	numConcurrentCalls                    int
+}
+
+func (s *SleepingService) listZoneRecommendations(project string, location string, recommenderID string) []*recommender.GoogleCloudRecommenderV1Recommendation {
+	d, err := time.ParseDuration(location)
```
minorczyka, line 59: That's very hacky :smile:

pkg/automation/listrecommendations_test.go
```diff
@@ -0,0 +1,120 @@
...
+		zones = append(zones, fmt.Sprint(duration))
+	}
+	mock := &SleepingService{zones: zones, numConcurrentCalls: 2}
+	result := ListRecommendations(mock, "", "")
+
+	assert.Equal(t, 0, len(result), "No recommedations expected")
+	assert.Equal(t, len(zones), len(mock.requestsDurations), "Wrong number of calls to listZoneRecommendations")
+	assert.True(t, sort.SliceIsSorted(mock.requestsDurations, func(i, j int) bool {
+		return mock.requestsDurations[i] < mock.requestsDurations[j]
+	}), "Faster goroutine should have finished first")
```
minorczyka, line 89: It's something that you really don't want to test. Depending on any thread sleeping in unit tests is a bad practice that can lead to flaky tests. Previous test is enough, we just check that we called everything we wanted and we don't have any deadlock.

pkg/automation/listrecommendations.go
```diff
@@ -0,0 +1,127 @@
...
+	listZoneRecommendations(project string, location string, recommenderID string) []*recommender.GoogleCloudRecommenderV1Recommendation
+
+	listZonesNames(project string) []string
+
+	getNumConcurrentCalls() int
+}
+
+// ZonesRecommendationsService implements Service interface for Recommender and Compute APIs,
+// using projects.locations.recommenders.recommendations/list and zones/list methods.
+type ZonesRecommendationsService struct {
```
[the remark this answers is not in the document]
ViktoriiaKovalova: I get the following linter error because of googleService is unexported:
pkg/automation/service.go:29:45: exported func NewGoogleService returns unexported type *automation.googleService, which can be annoying to use

pkg/automation/listrecommendations_test.go
```diff
@@ -0,0 +1,67 @@
...
+		assert.Equal(t, len(zones), mock.numberOfTimesListRecommendationsCalls, "Wrong number of listZoneRecommendaions calls")
+		assert.ElementsMatch(t, mock.zones, mock.zonesCalled, "listZoneRecommendations was called for different zones")
+	}
+}
+
+type BenchmarkService struct {
+}
+
+func (s *BenchmarkService) ListRecommendations(project string, location string, recommenderID string) []*recommender.GoogleCloudRecommenderV1Recommendation {
+	time.Sleep(time.Second)
```
minorczyka, line 48: If you really want to sleep in tests (e.g. to switch goroutines) than do it for much shorter time like a few milliseconds.

You don't want to have a single test running for a few second. You can easily imagine what would happen when we would have hundreds of tests and all of the would be run serially and most of them would take a few seconds.

pkg/automation/listrecommendations_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (s *BenchmarkService) ListZonesNames(project string) []string {
+	zones := []string{}
+	for i := 0; i < 100; i++ {
+		zones = append(zones, fmt.Sprintf("zone %d", i))
+	}
+	return zones
+}
+
+func BenchmarkGoroutines(b *testing.B) {
```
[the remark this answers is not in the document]
ViktoriiaKovalova: It's supposed to be a benchmark https://golang.org/pkg/testing/#hdr-Benchmarks not a test. They can be run by `go test` with `-bench=.` flag. Initially I just want to benchmark calls to the Google API, but then realised that relying on Internet connection in tests is probably a bad idea :)

pkg/automation/listrecommendations_test.go
ViktoriiaKovalova: I changed but it's not really a test, since it is not run by simple `go run` and I just thought that real calls to API might take some time, so chose 1 second as some approximation.

pkg/automation/listrecommendations_test.go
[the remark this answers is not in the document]
minorczyka: I didn't even know that something like that exists in go. So let's leave it as it is.

pkg/automation/listrecommendations_test.go
minorczyka: Now I understand that it's not a test, you can leave it as it is.

PR #5, opened 2020-07-02
add stopping and changing VMs

ViktoriiaKovalova:


ViktoriiaKovalova:
@minorczyka did you mean to use this embedding https://www.geeksforgeeks.org/embedding-interfaces-in-golang/ on the meeting? cause I'm not sure if that's it

minorczyka:
> @minorczyka did you mean to use this embedding https://www.geeksforgeeks.org/embedding-interfaces-in-golang/ on the meeting? cause I'm not sure if that's it

I meant that you can embed interfaces in structs: https://golang-examples.tumblr.com/post/132927916289/embedded-interface

ViktoriiaKovalova:
I changed the return types of ChangeMachineType and StopInstance so that would return error rather than crash immediately. I should probably do the same for list recommendations?

minorczyka:
> I changed the return types of ChangeMachineType and StopInstance so that would return error rather than crash immediately. I should probably do the same for list recommendations?

Yes you can do it right now.

minorczyka:
If you could also add license headers to all files, I forgot about it in the previous PR and I think we're supposed to have them.

pkg/automation/listrecommendations_test.go
```diff
@@ -7,54 +23,113 @@ import (
...
+	errorZone string
+	mutex     sync.Mutex
+	zones     []string
+}
+
+func (s *ErrorRecommendationService) ListZonesNames(project string) ([]string, error) {
+	return s.zones, nil
+}
+
+func (s *ErrorRecommendationService) ListRecommendations(project string, location string, recommenderID string) ([]*gcloudRecommendation, error) {
```
minorczyka, line 89: It would be nice to count how many times it was called. It would catch your bug with `break` inside the `select` statement.

pkg/automation/listrecommendations.go
```diff
@@ -58,12 +81,26 @@ func ListRecommendations(service GoogleService, project string, recommenderID st
...
+	errors := make(chan error)
+	quit := make(chan bool)
+
 	for i := 0; i < numWorkers; i++ {
 		go func() {
 			for zone := range zonesJobs {
-				results <- service.ListRecommendations(project, zone, recommenderID)
+				select {
+				case <-quit:
+					break
```
minorczyka, line 94: This `break` won't terminate execution of your `for` loop but your `select` https://golang.org/ref/spec#Break_statements

pkg/automation/listrecommendations.go
```diff
@@ -58,12 +81,26 @@ func ListRecommendations(service GoogleService, project string, recommenderID st
 		numWorkers = defaultNumWorkers
 	}
 
-	results := make(chan []*recommender.GoogleCloudRecommenderV1Recommendation, numberOfZones)
+	results := make(chan []*gcloudRecommendation, numberOfZones)
 	zonesJobs := make(chan string, numberOfZones)
+	errors := make(chan error)
```
minorczyka, line 86: I wouldn't bother with stopping other goroutines when we receive an error from one of the requests. We can just make all requests and just afterwards check if all of them succeeded.

I would introduce new type
```
type Result struct {
  recommendations []*gcloudRecommendation
  err error
}
```
And I would make `results := make(chan Result)`.

pkg/automation/listrecommendations_test.go
```diff
@@ -7,54 +23,113 @@ import (
...
+	s.mutex.Lock()
```
minorczyka, line 90: You don't need this mutex here. But you will need it if you add counting of this calls.

PR #7, opened 2020-07-16
Enable Vuetify

s17k:
- Added vuetify support, using `vue add vuetify`.

- Added the missing 'vuetify' type in tsconfig.json (this is a common problem, the related Github issue is linked in the second commit).

Now, vuetify will be installed automatically along with npm (this has been tested: I have cloned the repo into a clean directory and ran `npm install`, `npm run serve` which successfully loaded the Vuetify quickstart page)


s17k:
After reading more about this issue on Github, it seems that there is a cleaner solution to the problem with 'vuetify' having a missing type. I am therefore closing this PR and I will make a new one with the other solution proposed there.

s17k:
After reading even more, I realised that this is actually the recommended solution in Vuetify Docs [link]!(https://vuetifyjs.com/en/introduction/frequently-asked-questions/#when-adding-typescript-error-could-not-find-a-declaration-file-for-module-vuetify-lib)
Therefore, I am reopening this PR.

PR #12, opened 2020-07-16
Frontend build workflow

s17k:
**Added automatic builds, tests for frontend**.

I have chosen to only support the latest version of Node.js, as we are installing it ourselves anyway. 
I have fixed the sample test and the HelloWorld component so that they compile without warnings and so that the test passes.

I have also made sure that: [ failing tests cause the build to fail](https://github.com/googleinterns/recomator/pull/14).

Linter warnings now fail the build.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,29 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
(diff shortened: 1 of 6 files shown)


s17k:
Maybe adding `eslint "**/*.{vue,ts}" --max-warnings=0` would not be such a terrible solution after all.
I will add it in a moment.

s17k:
@minorczyka Where do you see the three builds exactly ?

minorczyka:
In the "Checks" section: https://github.com/googleinterns/recomator/pull/12/checks

It looks like the top build is for the recent source version and the bottom ones are for the previous versions. I guess...

PR #20, opened 2020-07-17
Frontend clean

s17k:
Added licenses to all source code except configs.
I also deleted all the quickstart code that I think will not be useful in the future.

Class-style components are not currently used, but this is fixed in branch frontend-clean-csc which will be merged next.


frontend/src/App.vue
```diff
@@ -1,57 +1,45 @@
...
-        <span class="mr-2">Latest Release</span>
-        <v-icon>mdi-open-in-new</v-icon>
-      </v-btn>
-    </v-app-bar>
-
-    <v-content>
-      <HelloWorld />
-    </v-content>
-  </v-app>
+  <!-- Copyright 2020 Google LLC
```
minorczyka, line 2: According to https://stackoverflow.com/questions/41228833/how-to-comment-code-in-a-vue-js-file you should be able to put these comments outside the template tag.

s17k:
I thought so at the beginning as well, but removing it didn't break anything. I will take another look.

s17k:
I reverted deleting `favicon.ico` now by mistake, I will remove it again.

PR #30, opened 2020-07-20
Frontend mock

s17k:
A new page (/mock) with fake data that shows how the actual app might look like.


s17k:
@minorczyka Do we ever want to merge this to master ? I would argue that we shouldn't, as this will shortly be outdated anyway.

minorczyka:
I think we shouldn't merge this mock branch into master. 

But I'll leave decision up to you, if it will be easier for you to have it on the master then I won't oppose.

PR #35, opened 2020-07-21
Closes #27. Adds a pre-commit hook for frontend.

### LINKED ISSUE #27: Add a git hook that makes sure that you don't commit/push changes that have linter warnings.

Podkamienna:


frontend/hooks/pre-commit
```diff
@@ -0,0 +1,34 @@
...
+targetdir=${tmpdir}recomator/frontend
+
+cd $targetdir
+
+format_file() {
+  file="${1}"
+  a=${file#*frontend/}
+  if [ -f $a ]; then
+    npm run lint $a
+    git add $a
```
s17k, line 28: I don't really understand what is the role of `git add $a` here. ? Could you please explain what was your motivation ?
I have made a simple test and `git diff-index --cached `only shows files that have already been added using `git add`.

minorczyka:
There's also some issue with your GitHub account, have you successfully added your @google email to it?

frontend/hooks/pre-commit
s17k: Oh, I think I got it. `npm run lint`auto-fixes everything it touches, so you need to stage these changes.

Podkamienna: So actually it only fixes the file, which name is in the variable a, so it is possible that the file will be changed after that command. So if there was no git add later, the changes made by the linter would not be staged for commit.

Podkamienna:
@googlebot I signed it!

Podkamienna:
@minorczyka I was sure I did, but my account wasn't registered in the google base of github accounts, so I probably did something wrong.

frontend/hooks/pre-commit
s17k: This might cause someone a minor headache if they were planning to only commit the changes done before `git add` and not the ones made after. I personally never do that, but I am not sure if other people do.

minorczyka: Maybe we can just use some pre-made solution instead of inventing something on our own? I found some article about it: https://dev.to/juliang/configuring-prettier-and-typescript-compiler-as-a-pre-commit-hook-44jh But I don't have any experience with any of that tools.

PR #36, opened 2020-07-21
Interfaces for frontend data model added

### LINKED ISSUE #31: Create a model and types for recommendations

s17k:
It is not used by any piece of code at the moment, so I think it is fine not to test it before merging with master. We can test the code that actually uses this, once it is written.
Closes #31.


s17k:
I am closing this as I have changed a thing from here in a different branch that already has the only commit from here.

PR #37, opened 2020-07-22
Frontend store logic skeleton

s17k:
- Added the store and the Recommendations module (they are separated already so that they can be individually imported for tests)
- Added the data model for Recommendation (following the Recommender API: https://cloud.google.com/recommender/docs/reference/rest/v1beta1/projects.locations.recommenders.recommendations#Recommendation)
- I re-used the sample recommendation from Viktoriia's PR for tests.
- As we are already using class-style components for Vue, I decided to do the same for Vuex using  [vuex-module-decorators](https://github.com/championswimmer/vuex-module-decorators). It has the MIT license so all is good.

I think we should merge this now (as long as no critical problems are found), so that Alicja and me can build upon it together. If we were to work on this branch for a bit longer, the PR would become too big.


frontend/src/data-model/model.ts
```diff
@@ -0,0 +1,66 @@
+/* Copyright 2020 Google LLC
```
minorczyka, line 1: Let's just put this file into the store directory.

frontend/src/store/recommendations.ts
```diff
@@ -0,0 +1,80 @@
...
+import "@/data-model/model.ts";
+
+import { Module, VuexModule, Mutation, Action } from "vuex-module-decorators";
+
+@Module
+export default class extends VuexModule {
+  recommendations: Record<string, Recommendation> = {};
+
+  @Mutation
+  ADD_RECOMMENDATION(recommendation: Recommendation) {
```
minorczyka, line 24: Let's be consistent and just use camelCase.

frontend/src/store/recommendations.ts
```diff
@@ -0,0 +1,80 @@
...
+  getRecommendationProject(recommendationName: string): string {
+    return "rightsizer-test";
+  }
+
+  @Action
+  getRecommendationResourceName(recommendationName: string): string {
+    return "timus-test-for-probers-n2-std-4-idling";
+  }*/
+
+  @Action
```
minorczyka, line 46: I think you want to have a simple getters here instead of actions. Actions are for performing some asynchronous operations.

frontend/src/store/recommendations.ts
```diff
@@ -0,0 +1,80 @@
...
+    console.assert(
+      !(recommendation.name in this.recommendations),
+      "the key %s is present in the store already",
+      recommendation.name
+    );
+    this.recommendations[recommendation.name] = recommendation;
+  }
+
+  /*
+  TODO: To be implemented
```
minorczyka, line 35: Instead of implementing all these `getRecommendation[...]` getters I would just implement `getRecommendation(name): Recommendation` function. Most of this getters that just access some fields of the recommendation would become unnecessary in that case. And a little bit more complicated getters can be just implemented as a simple functions, not related with the store.

frontend/tests/unit/example.spec.ts
```diff
@@ -12,15 +12,67 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */
 
-import { shallowMount } from "@vue/test-utils";
```
minorczyka, line 15: Please change the name of this file to something meaningful.

frontend/tests/unit/example.spec.ts
```diff
@@ -12,15 +12,67 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+  },
+  recommenderSubtype: "CHANGE_MACHINE_TYPE",
+  stateInfo: {
+    state: "CLAIMED"
+  }
+};
+
+describe("Store tests", () => {
+  test("ADD_RECOMMENDATION", () => {
+    const store = new Vuex.Store({
```
minorczyka, line 65: You should be able to just import store from the `frontend/src/store/index.ts`. Then you won't have to create it each time and you won't have to import Vue and Vuex in this test.

frontend/tests/unit/example.spec.ts
```diff
@@ -12,15 +12,67 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      modules: {
+        recommendationsStore: RecommendationStore
+      }
     });
-    expect(wrapper.text()).toContain(msg);
+
+    store.commit("ADD_RECOMMENDATION", sampleRecommendation);
+    // For some very weird reason, using expect().toHaveProperty() only works if the name is short
+    expect(
```
minorczyka, line 73: In such cases we should also check if recommendation in store is equal to what we inserted there.

frontend/tests/unit/example.spec.ts
s17k: Isn't there only one instance of the store in frontend/src/store/index.ts ? I would like to have fresh stores for each test.

frontend/tests/unit/example.spec.ts
s17k: Yes, this is more of a skeleton of a test than a real test. Do you want me to expand it here ?

frontend/tests/unit/example.spec.ts
minorczyka: Oh you're right, it's exporting object instance so it will be most probably reused. I don't know how this JavaScript unit tests work, maybe this instance will be recreated for every test run but most probably not. You can check if you want or you can just ignore this comment.

frontend/tests/unit/example.spec.ts
s17k: I totally missed that, thanks!

frontend/src/store/recommendations.ts
s17k: getters don't have arguments unfortunately.
As was mentioned elsewhere, I will move Recommendation-object related functions outside of the store and only keep the method that fetches the recommendation by name.

frontend/src/store/recommendations.ts
s17k: I was following a convention to name mutations in all caps (this would help differentiate between them and actions and not accidentally call a mutation like you would an action - without a commit). Let me know if you think that is okay.

s17k:
> I didn't like `vuex-module-decorators` and that's why I haven't suggested it to you. In my opinion, it doesn't make store more type-safe, readable or less verbose. But feel free to use it if you like it.

I am not completely sure either, the code does look more readable now but I will take reconsider once there is more of it. I will leave it for now, so that we can merge this and I will let this be decided later.

frontend/src/store/recommendations.ts
minorczyka: There's a way to do it with getters but it's not straightforward, you just have to return function with parameter from getter: https://vuex.vuejs.org/guide/getters.html#method-style-access

It's also one of the reasons why I didn't like this typescript library for stores, it looks weird with parameterized getters.

I think that using actions as getters will be counter intuitive for new people reading this code, in the end action's description in the docs (https://vuex.vuejs.org/guide/actions.html) clearly says what's the purpose of actions.

frontend/src/store/recommendations.ts
minorczyka: If we want to use https://github.com/ktsn/vuex-class with class based components than I would argue that it won't be really necessary and will look weird.

What you describe is a convention from React (more specifically Flux and Redux). If you want we can choose this approach and go even farther with declaring constants for names of mutations: https://vuex.vuejs.org/guide/mutations.html#using-constants-for-mutation-types

frontend/src/store/recommendations.ts
s17k: I have just moved these somewhere else as simple functions, they were overwritten temporarily but it should be all good now.

frontend/src/store/recommendations.ts
```diff
@@ -0,0 +1,45 @@
...
+    // prevent overwriting
+    console.assert(
+      !(recommendation.name in this.recommendations),
+      `recommendation name ${recommendation.name} is present in the store already`
+    );
+    this.recommendations[recommendation.name] = recommendation;
+  }
+
+  @Action
+  getRecommendation(recommendationName: string): Recommendation {
```
minorczyka, line 33: When I think about it right now, I'm not even sure if we will need getting recommendation based on name. It's very likely that we will need all recommendations (or one page of recommendations). We can leave it as it is for now but most probably we will change it.

PR #44, opened 2020-07-24
Frontend fetch recommendations

Podkamienna:
Adds a method to the store for fetching recommendations.


frontend/src/views/Home.vue
```diff
@@ -25,6 +25,18 @@ limitations under the License. -->
 // @ is an alias to /src
 import { Component, Vue } from "vue-property-decorator";
 import HelloWorld from "@/components/HelloWorld.vue";
+import RecommendationStore from "@/store/recommendations";
```
minorczyka, line 28: Please do not commit code that you used for testing.

frontend/src/store/recommendations.ts
```diff
@@ -39,7 +43,23 @@ export default class extends VuexModule {
   }
 
   @Action
-  fetchRecommendations() {
-    //TODO
+  async fetchRecommendations() {
```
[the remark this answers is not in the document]
Podkamienna: Why would that be a problem? 

Is it possible, that the sequence of the operations below would happen? Or is it something else?
1. The first fetch is called
2. The second fetch is called
3. They both receive the recommendations
4. They start updating recommendations
5. They both try to add a recommendation with name "xyz" but different contents
6. The first checks if the name is present
7. The second checks if the name is present
8. The first adds a name
9. The second overwrites the recommendation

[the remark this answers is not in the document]
s17k: I think that is one of the possible problematic scenarios. Even if the names didn't overlap, we still wouldn't want to display two states of generated recommendations at once.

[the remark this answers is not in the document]
Podkamienna: I tried to dispatch second fetchRecommendations while the first was in progress. I logged to the console the progress, and it looked as if the second one was dispatched after the first one ended. I also added another action, which I dispatched between the two dispatches of fetchRecommendations and it was executed in the middle of the first dispatch of fetchRecommendations.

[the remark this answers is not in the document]
Podkamienna: Also, do we want to allow fetching recommendations twice or not? If yes, then there is a case, when one "thread" finished fetching recommendations, but did not free the mutex. And the only solution that I see for that issue is that we would need a mutex to be on fetching and checking if we have finished. Because there is sleep executed in each rotation of the loop, then it would probably not hurt the performance. But I am not sure. Is that solution ok?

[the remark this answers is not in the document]
minorczyka: I just imagine that running two fetching operations at the same time can lead to some very subtle bugs and I just think it's simpler to prevent such situation at all. 

One potential problem that just came to my mind:
- first fetch loaded list of recommendations,
- user applies one recommendation,
- we display applying progress,
- second fetch loaded list of recommendations,
- overriding of recommendations caused disappearance of applying progress indicator.

After a short time applying indicator will be probably visible again but it's not a good user experience and in worst case (depending on implementation details) progress indicator might not even reappear at all.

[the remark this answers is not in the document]
Podkamienna: I am not sure if I understand correctly, but when the status of the recommendation would be incorrect because of the second fetch of the same resource it would also be incorrect after reloading the page. So maybe it would be better to fetch the status of each loaded recommendation, while executing the fetch? But it may be really slow if there are many recommendations.

Anyway, we want to be able to call fetch only once?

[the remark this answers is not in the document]
s17k: I totally agree that allowing two fetching operations at the same time does not sound too good.
But, I think it would be easier to just prevent this from happening in the first place. For now, we only plan to fetch recommendations when the application starts anyway, right ? Even if we added a refresh button, couldn't we just disable it for as long as the previous fetch is running ? This way we could avoid doing mutexes and all that stuff.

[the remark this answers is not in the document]
minorczyka: I would just add some information to our store telling if fetch is currently in progress. fetchRecommendations would then check it at the very beginning and if other fetch was already in progress, it would immediately return.

I'm not sure why are you talking about threads and mutexes. JavaScript doesn't support parallel execution (without web workers), everything is executed on a single thread. You can read some basics about JavaScript concurrent model here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop

frontend/src/store/recommendations.ts
```diff
@@ -29,6 +36,26 @@ export default class extends VuexModule {
     this.recommendations[recommendation.name] = recommendation;
   }
 
+  @Mutation // Only one fetching may be in progress at once
+  tryStartFetching(result: ReferenceWrapper<boolean>) {
+    if (this.inProgress) {
+      result.setValue(true);
+    }
```
minorczyka, line 43: It's always overwritten in the next line.

frontend/src/store/recommendations.ts
```diff
@@ -39,7 +66,39 @@ export default class extends VuexModule {
...
   @Action
-  fetchRecommendations() {
-    //TODO
+  async fetchRecommendations() {
+    const isAnotherFetchInProgress = new ReferenceWrapper(false);
+    this.context.commit("tryStartFetching", isAnotherFetchInProgress);
+
+    if (isAnotherFetchInProgress.getValue()) {
+      return;
+    }
```
minorczyka, line 75: In fact, is it any better than directly checking `state.inProgress` in this place?

frontend/src/store/recommendations.ts
```diff
@@ -39,7 +66,39 @@ export default class extends VuexModule {
...
+
+    let response;
+    let responseJson;
+
+    for (;;) {
+      response = await fetch(`${SERVER_ADDRESS}/recommendations`);
```
minorczyka, line 81: Currently it's not implemented in the fake service (I created an issue) but we should also handle errors returned by this endpoint. You can assume that it will also have an "errorMessage" field  and will return a status code different than 200.

frontend/src/store/recommendations.ts
```diff
@@ -14,10 +14,17 @@ limitations under the License. */
...
+import { ReferenceWrapper } from "./reference";
+
+const SERVER_ADDRESS: string = getServerAddress();
+const REQUEST_DELAY = 100;
 
 @Module
 export default class extends VuexModule {
   recommendations: Record<string, Recommendation> = {};
+  progress = 0; // percent of recommendations loaded
+  inProgress = false;
```
minorczyka, line 27: Just an idea: we can make `progress: number | null = null` and then we can tell if it's in progress by checking `progress != null`

In fact, we could even introduce some utility type that can be used to tell if operation is still in progress, what's the progress and if it succeeded or what's the error message. But I'm not sure if it can be easily done in TypeScript and if it will be convenient to use.

frontend/src/store/recommendations.ts
```diff
@@ -39,7 +66,39 @@ export default class extends VuexModule {
   }
 
   @Action
-  fetchRecommendations() {
-    //TODO
+  async fetchRecommendations() {
```
minorczyka, line 69: We should be able to query different pages of recommendations list and reuse `pageToken` for it. Currently we don't even extract `pageToken` from the results.

frontend/src/views/Home.vue
```diff
@@ -25,6 +25,9 @@ limitations under the License. -->
 // @ is an alias to /src
 import { Component, Vue } from "vue-property-decorator";
 import HelloWorld from "@/components/HelloWorld.vue";
+import Vuex from "vuex";
```
s17k, line 28: We already have that in store/index.ts. You only need to register Vuex once.

frontend/src/store/recommendations.ts
Podkamienna: Yes, because if we would be in the first of the two fetch calls, then `state.inProgress` would also be set to true.

frontend/src/store/recommendations.ts
Podkamienna: I can add it, but I don't understand right now, what will we be using it for. Will we want to call fetch every n seconds just to have updated results? Or will we ask user for it when they reload the page? Or are we just going to save it in localstorage and use if it's there?

frontend/src/store/recommendations.ts
```diff
@@ -38,25 +40,34 @@ export default class extends VuexModule {
...
     this.progress = progress;
   }
 
+  @Mutation
+  setError(errorCode: number, errorMessage: string) {
+    this.errorCode = errorCode;
+    this.errorMessage = errorMessage;
+
+    console.log(errorMessage);
+    console.log(errorCode);
```
minorczyka, line 68: Please do not commit these statements.

frontend/src/store/recommendations.ts
minorczyka: Let's leave pagination for later maybe we even won't use it at all.

frontend/src/store/recommendations.ts
minorczyka: I mean, is there any benefit of doing it this way instead of:
```
 const isAnotherFetchInProgress = this.progress !== null;
 this.context.commit("tryStartFetching");
```
Or why not just replace all of it with
```
if (this.progress !== null) {
  return
}
this.context.commit('setProgress', 0);
```

frontend/src/store/recommendations.ts
Podkamienna: Ok, I just thought, that it may make sense, to print those if we don't have the function to check for errors yet (to avoid being confused, when receiving errors).

frontend/src/store/recommendations.ts
Podkamienna: Okay, so I thought that it made a difference (calling a function vs writing statements in code), but probably it doesn't.

PR #62, opened 2020-07-30
Update model.ts

a-piasta:
Komentarze powinny zaczynać się wielką literą i kończyć kropką.


s17k:
It is great that you want to contribute. I don't think I agree with this change, though. I understand that this is a convention that you like, but I don't think it has a significant impact on project objectives and caring about it has a negative one. I am closing this PR.

PR #65, opened 2020-07-31
Frontend component development

s17k:
- Filters/search row implemented (one thing that wasn't included in the mock is disregarding the case of the search and cell text, so that searching for 'save' yields 'Save money...')
- Filters test (currently only resource names are tested) ; the compiler on the Github build had a different opinion about the type of `CoreTable.prototype.constructor` (it worked locally), so I cast it as 'any' and ignored the related rule in .eslintrc.js
- Core table components structure is now present
- Implemented (the templates for) Resource name, Project name, Recommender Subtype, Description columns
- optional execution of fetchRecommendations (fake-service integration) at start (uncomment a single line)
- remembering properties used for sorting/grouping (such as project/type or resource name) so that v-data-table can use them.

**Before:**
Screenshot 2020-07-30 at 5 24 11 PM
**After:**
Screenshot from 2020-08-04 18-53-08


s17k:
Just to let you know, I will not make more commits here (except for implementing fixes/suggestions from the reviews).

frontend/src/components/CoreTable.vue
```diff
@@ -0,0 +1,189 @@
...
+      text: "Savings/cost per week",
+      value: "costCol",
+      groupable: false,
+      sortable: true
+    },
+    { text: "", value: "applyAndStatus", groupable: false, sortable: false }
+  ];
+
+  get filteredRecommendations(): Recommendation[] {
+    const rootStoreState = this.$store.state as IRootStoreState;
```
minorczyka, line 128: I would recommend mapping `recommendations` from the `store` with the `State` annotation from `vuex-class`. Then you can have a property in this class `@State recommendations!: Recommendation[]` and you can just access it with `this.recommendations` and it will be typesafe.

(I might've put `!` in wrong place and you probably have to use path to your module `recommendationsStore` you can find details here: https://github.com/ktsn/vuex-class)

frontend/src/components/CoreTable.vue
```diff
@@ -0,0 +1,189 @@
...
+        )
+    );
+  }
+
+  get filteredRecommendationsWithExtras(): Recommendation[] {
+    return this.filteredRecommendations.map(
+      recommendation => new RecommendationExtra(recommendation)
+    );
+  }
+
```
minorczyka, line 143: I would put those getters into the store. Thanks to that, we will keep all business logic in one place and it should be easier to test (I only guess but I think that store is easier to test than components).

frontend/src/components/CoreTable.vue
```diff
@@ -0,0 +1,189 @@
...
+  // TODO: grouping should temporarily increase items shown to all
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  onGroupByUpdated(groupByCategories: string[]): void {
+    // TODO: update itemsPerPage
+  }
+
+  // TODO: once there is a new non-empty groupBy, close (toggle) all opened projects/types
+
+  static isRecommendationInResults(
```
minorczyka, line 154: It can be just a normal function outside this class.

frontend/src/components/DescriptionCell.vue
```diff
@@ -0,0 +1,34 @@
...
+  <td>
+    {{ description() }}
+  </td>
+</template>
+<script lang="ts">
+import Vue from "vue";
+import { Component } from "vue-property-decorator";
+import { getRecomendationDescription } from "../store/model";
+
+const DescriptionCellProps = Vue.extend({
```
minorczyka, line 24: I would just use `Prop` annotation from `vue-class-component` for defining props (it's shorter and cleaner and you can specify their type).

frontend/src/components/FiltersRow.vue
```diff
@@ -0,0 +1,141 @@
...
+    </td>
+  </tr>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from "vue-property-decorator";
+import { IRootStoreState } from "../store/root";
+
+@Component
+export default class FiltersRow extends Vue {
```
minorczyka, line 73: You can map (probably) everything in this class with annotations.

frontend/src/store/model.ts
```diff
@@ -93,33 +94,70 @@ export function getRecommendationProject(
...
   );
-  // example units: "-73"
-  const cost = parseInt(recommendation.primaryImpact.costProjection.cost.units);
+  // example units: "-73", example nanos: 4200000000 => -73.42
+  // Sometimes we will only get the 'nanos' field
+  //  (only observed for snaphshots for now)
+
+  let cost = 0;
+
+  if (Object.prototype.hasOwnProperty.call(costObject, "units"))
```
minorczyka, line 126: You can just use `costObject.units !== undefined`.

frontend/src/store/recommendations.ts
```diff
@@ -99,10 +100,23 @@ const actions: ActionTree<IRecommendationsStoreState, IRootStoreState> = {
...
+      context.commit("addRecommendation", recommendation);
     }
 
-    context.commit("recommendationsStore/endFetching");
+    context.commit("endFetching");
+  }
+};
+
+const getters: GetterTree<IRecommendationsStoreState, IRootStoreState> = {
+  allProjects(state): Array<string> {
```
minorczyka, line 111: nit: `string[]` is shorter and probably nicer looking but both are correct.

frontend/src/components/DescriptionCell.vue
s17k: This might have been discontinued: 

- Class-style-component docs suggest my approach: [link](https://class-component.vuejs.org/guide/props-definition.html)

- But, interestingly, the Vue docs suggest using PropType (type, not decorator) from 'vue': [link](https://vuejs.org/v2/guide/typescript.html#Annotating-Props)

From what I also saw [here](https://frontendsociety.com/using-a-typescript-interfaces-and-types-as-a-prop-type-in-vuejs-508ab3f83480) this does seem to have been introduced recently.

I think the second option is the best as it is the least verbose one, the recommended one and is type-safe.
Let me know what you think.

frontend/src/components/CoreTable.vue
s17k: I tried doing that and I couldn't import it as it seems that .Vue files only export the main class.
But, I could just put in the core_table store module, which I didn't think of earlier! I will do that.

frontend/src/components/DescriptionCell.vue
minorczyka: Sorry I meant from `vue-property-decorator`.

s17k: What do you think about using PropType instead ? We wouldn't need to add an additional dependency, this is the solution suggested on the Vue page and it is just as verbose ?

frontend/src/components/CoreTable.vue
```diff
@@ -0,0 +1,189 @@
...
+<template>
+  <v-data-table
+    dense
+    :headers="headers"
+    :items="filteredRecommendationsWithExtras"
+    show-group-by
+    v-on:update:group-by="onGroupByUpdated"
+    :items-per-page="itemsPerPage"
+    :single-select="false"
+    v-model="$store.state.recommendationsStore.selected"
```
s17k, line 23: This is really bad, it is modifying the store state without mutations... :facepalm:

frontend/src/components/CoreTable.vue
s17k: This code belongs in the store anyway, so this will not be needed in this exact place. I will keep that in mind for some future problems like that, though.

frontend/src/components/FiltersRow.vue
s17k: Using vuex-class decorators could be useful for getters, but I don't really see what to do with the setters ? What did you have in mind ?

s17k:
Sorry. I forgot to stage a few files in the merge, but you can hopefully see that nothing tragic happened by looking at the commit hashes and I just removed the last commit.

frontend/src/components/DescriptionCell.vue
minorczyka: If we do this, does autosuggestion and typechecking work when you use e.g. `this.rowRecommendation`?

frontend/src/components/FiltersRow.vue
minorczyka: I haven't noticed that you're using it in that `v-model`. That's unfortunate but I agree that in that case there's not much we can do better here.

frontend/src/store/root.ts
```diff
@@ -28,13 +35,28 @@ export interface IRootStoreState {
   // Therefore, the ! operator needs to be used whenever the state of any module
   //  is accessed from outside.
   recommendationsStore?: IRecommendationsStoreState;
+  coreTableStore?: ICoreTableStoreState;
 }
 
+const getters: GetterTree<IRootStoreState, IRootStoreState> = {
```
minorczyka, line 41: Why have you put it here and not into one of the modules?

frontend/tests/unit/core_table.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+      // deep copy
+      JSON.stringify(sampleRecommendation)
+    ) as Recommendation;
+
+    // Check if we accept resoruce name: bob-vm0 when searching for bob
+    fakeStore.commit("coreTableStore/setResourceNameSearchText", "bob");
+    freshSampleRecommendation.content.operationGroups[0].operations[0].resource =
+      "//compute.googleapis.com/projects/search/zones/us-east1-b/instances/bob-vm0";
+    expect(
+      isRecommendationInResults(
```
minorczyka, line 33: You should also test `isRecommendationInResults` function itself. You rely on it in your business logic and in your unit tests here so if this function will be buggy these tests might not catch it.

frontend/src/store/root.ts
s17k: Because the getter is using two different modules.

frontend/src/components/DescriptionCell.vue
s17k: I checked it earlier and the types are enforced, you can see it in the autocomplete and I even read the Vue source code where you can see that the types are checked at runtime (so it will be caught if the prop passed has a different type) [link](https://github.com/vuejs/vue/blob/dev/src/core/util/props.js#L146)

frontend/tests/unit/core_table.spec.ts
s17k: Yes, thanks!

s17k:
I swear I will try to make future PRs smaller than this 😆

frontend/src/components/DescriptionCell.vue
minorczyka: That's great, feel free to use it.

frontend/src/store/root.ts
minorczyka: You can access it with `rootGetters` https://stackoverflow.com/questions/45999623/how-to-access-the-getter-from-another-vuex-module

s17k:
@minorczyka Not a problem, but this would mean that I would need to extend the getter to take rootState as a parameter as well (which is indeed possible), so as to get access to the CoreTable store through the root. Just to confirm, is that what you wanted ?

Not sure if that is the motivation for this suggestion, but it does bug me that the store name "recommendations" is practically meaningless in this application, it maybe loosely implies the actual data is stored there. I deffinitely plan to move "selected" (the array with selected recommendation) to the core table store, and maybe rename it as external_data (as in, it came from outside the app).

frontend/src/store/root.ts
s17k: That is fine. Could you please explain why do you think that is a good idea ? Do you want to unclutter root.ts perhaps ? I originally opted for putting it there so that we can keep modules independent (no calls to sibling/root modules) and so that it can be assumed that you only need to mock one module when you are testing it.

minorczyka: I thought about the root store as just a place where you configure other submodules. It didn't feet there for me. But maybe you're right. This getter probably also doesn't fit into recommendations store and making more independent modules is better. 

You can leave it as it is, I don't have any strong opinion about it.

PR #68, opened 2020-08-03
Added a test for fetch action of recommendation store.

Podkamienna:
Test is done using a package from npm that allows its user to substitute value returned by fetch.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -9,18 +9,19 @@
     "lint": "vue-cli-service lint"
   },
   "dependencies": {
+    "@types/node": "^12.12.6",
     "babel-plugin-transform-decorators": "^6.24.1",
     "core-js": "^3.6.5",
     "vue": "^2.6.11",
-    "vue-class-component": "^7.2.3",
+    "vue-class-component": "^7.2.5",
     "vue-property-decorator": "^8.4.2",
     "vue-router": "^3.2.0",
-    "vuetify": "^2.2.11",
+    "vuetify": "^2.3.7",
     "vuex": "^3.4.0"
   },
   "devDependencies": {
     "@babel/plugin-transform-modules-commonjs": "^7.10.4",
-    "@types/jest": "^24.0.19",
+    "@types/jest": "^24.9.0",
     "@typescript-eslint/eslint-plugin": "^2.33.0",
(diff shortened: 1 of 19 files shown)


frontend/tests/unit/recommendations.spec.ts
```diff
@@ -12,6 +12,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */
 
+import { enableFetchMocks } from "jest-fetch-mock";
+enableFetchMocks();
+fetchMock.dontMock();
```
s17k, line 17: It would be nice to comment on what this does, considering it looks as if you are enabling something and then  .. disabling it just after that (I know you are not, but it would be nice to know what this does).

frontend/package.json
```diff
@@ -9,18 +9,19 @@
     "lint": "vue-cli-service lint"
   },
   "dependencies": {
+    "@types/node": "^12.12.6",
```
minorczyka, line 12: It probably should be in the devDependencies, but it doesn't make a big difference for us anyway.

frontend/tests/unit/recommendations.spec.ts
```diff
@@ -40,10 +44,90 @@ describe("Recommendation-type objects", () => {
...
+});
 
+describe("Fetching recommendations", () => {
   test("Getting the instance that the recommendation references", async () => {
     expect(
       Model.getRecommendationResourceShortName(sampleRecommendation)
     ).toEqual("alicja-test");
   });
+
+  test("Fetching works correctly when given response without errors", async () => {
```
minorczyka, line 56: It's just my personal opinion (we don't have any style guide to follow) but I think you're using too many empty lines in your code here.

In general, using many empty lines makes code harder to read (less code fits on the screen) but it's good to use empty lines to separate "sections" of code. In that case I would group:
- L57 and L59 (test options)
- L61-L75 (setting up fetch mock)
- L77-L81 (actions being tested)

It applies also to a few other places.

Podkamienna: I will try to change the way I use the empty lines. I was using so many of them, because it made it easier for me to remember the logic I used, when writing the code.

PR #73, opened 2020-08-05
Test operation

Podkamienna:
Added functions that should be called when test operation is received.


pkg/automation/test.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Checks if the machine type of the instance specified by given project, zone and instance
+// matches the given value or valueMatcher
+func (s *googleService) TestMachineType(project string, zone string, instance string, value interface{}, valueMatcher *gcloudValueMatcher) (bool, error) {
+	machineInstance, err := s.GetInstance(project, zone, instance)
+	if err != nil {
+		return false, err
+	}
+
+	machineType := machineInstance.MachineType
```
minorczyka, line 61: I would say that declaring this variable is unnecessary. It only makes code longer and a little bit harder to follow. I would just use `machineInstance.MachineType` directly.

pkg/automation/test.go
```diff
@@ -0,0 +1,78 @@
...
+	"google.golang.org/api/recommender/v1"
+)
+
+type gcloudValueMatcher = recommender.GoogleCloudRecommenderV1ValueMatcher
+
+// Checks if the string toTest matches the member MatchesPattern of valueMatcher
+// If valueMatcher is not nil. Otherwise, if value is not nil it is interpreted as string
+// And equality of value.(string) and toTest is checked. If both value and valueMatcher are nil,
+// then it results in an error
+func test(toTest string, value interface{}, valueMatcher *gcloudValueMatcher) (bool, error) {
```
minorczyka, line 31: Name of this function could be more descriptive. Plus word `test` is quite common and it might collide with some other symbols e.g in unit tests.

minorczyka: In fact, I would split this function into `testValue` and `testValueMatcher`, code would look cleaner and it would be easier to understand what they do.

pkg/automation/test.go
```diff
@@ -0,0 +1,78 @@
...
+	if !ok {
+		return false, typeError()
+	}
+
+	return valueString == toTest, nil
+}
+
+// Checks if the machine type of the instance specified by given project, zone and instance
+// matches the given value or valueMatcher
+func (s *googleService) TestMachineType(project string, zone string, instance string, value interface{}, valueMatcher *gcloudValueMatcher) (bool, error) {
```
minorczyka, line 55: When I've seen this function for the first time I didn't understand why we pass `value` and `valueMatcher` as parameters. I had to look at Recommender API library documentation to understand it. It would be nice to put some comments about it (probably even copy/paste or summary of comments from these fields in Recommender API library).

pkg/automation/test.go
```diff
@@ -0,0 +1,78 @@
...
+type gcloudValueMatcher = recommender.GoogleCloudRecommenderV1ValueMatcher
+
+// Checks if the string toTest matches the member MatchesPattern of valueMatcher
+// If valueMatcher is not nil. Otherwise, if value is not nil it is interpreted as string
+// And equality of value.(string) and toTest is checked. If both value and valueMatcher are nil,
+// then it results in an error
+func test(toTest string, value interface{}, valueMatcher *gcloudValueMatcher) (bool, error) {
+	if value == nil {
+		if valueMatcher == nil {
+			return false, operationUnsupportedError()
```
minorczyka, line 34: I'm not sure that it should return error in such case. If both `value` and `valueMatcher` are nil then there's nothing to check so it should return true.

pkg/automation/error.go
```diff
@@ -0,0 +1,31 @@
...
+limitations under the License.
+*/
+
+package automation
+
+import "errors"
+
+// Error returned when some type does not
+// fit its asserted type
+func typeError() error {
```
minorczyka, line 23: If we want to introduce generic errors that will be reused in our application, please read https://blog.golang.org/go1.13-errors

But in these functions that you introduced I would argue that it's not necessary. I would just create errors with specific messages there (e.g. "value parameter must have string type" or something similar). Thanks to that if someone has access to the codebase then it's much easier to find causes of errors, you can just easily search for specific error messages.

pkg/automation/test.go
Podkamienna: I had mixed feelings about this. Because on the one hand it would be great to be able to accept as much values as possible. But in the documentation it is written that either value or valueMatcher will be specified. So it surely means, that it shouldn't happen, that both are nil. I wasn't sure what about the case, when they both are specified - but if the case with two nils is not accepted, then nor should be the one with zero nils? And even if the second one should be accepted, both value and valueMatcher should be checked.

pkg/automation/test.go
```diff
@@ -0,0 +1,103 @@
...
+		return false, err
+	}
+
+	return r.MatchString(toTest), nil
+}
+
+// Checks if the string toTest matches the member MatchesPattern of valueMatcher,
+// if valueMatcher is not nil. Otherwise, if value is not nil it is interpreted as string
+// and equality of value.(string) and toTest is checked.
+func testMatching(toTest string, value interface{}, valueMatcher *gcloudValueMatcher) (bool, error) {
```
minorczyka, line 60: Can you write unit tests for this function?

PR #74, opened 2020-08-05
Snapshot name

Podkamienna:
Added creating the name of a snapshot. Changed the signature of the CreateSnapshot function.


pkg/automation/disks.go
```diff
@@ -16,12 +16,53 @@ limitations under the License.
 
 package automation
 
-import "google.golang.org/api/compute/v1"
+import (
+	"time"
+
+	"github.com/xyproto/randomstring"
```
minorczyka, line 22: This library doesn't look like well maintained and it should be really easy to write simple function generating random string (it doesn't have to be human friendly). So I wouldn't introduce unnecessary dependency.

pkg/automation/disks.go
```diff
@@ -16,12 +16,53 @@ limitations under the License.
...
+// Returns current time in the fromat YYYYMMDDHHMMSS
+func getTimestamp() string {
+	t := time.Now().UTC()
+	return t.Format("20060102150405")
+}
+
+// Returns the name of the snapshot, following
+// the convention described here:
+// https://cloud.google.com/compute/docs/disks/scheduled-snapshots#names_for_scheduled_snapshots
+func getSnapshotName(zone, disk string) string {
```
minorczyka, line 47: Please write som unit tests for this function. Especially I would like to have a test checking that length of returned name is always <= 63, no matter how long disk and zone names are.

minorczyka: And function name like "randomSnapshotName" would be also better, to indicate it's non-deterministic.

pkg/automation/disks.go
```diff
@@ -16,12 +16,53 @@ limitations under the License.
...
+	randomSequenceLen := maxSnapshotnameLen - len(result)
+	result += randomstring.HumanFriendlyString(randomSequenceLen)
+
+	return result
+}
 
 // CreateSnapshot calls the disks.createSnapshot method.
 // Requires compute.disks.createSnapshot or compute.snapshots.create permission.
-func (s *googleService) CreateSnapshot(project, zone, disk, name string) error {
+func (s *googleService) CreateSnapshot(project, zone, disk string) error {
```
minorczyka, line 63: I would leave this `name` parameter here. We can just use this `getSnapshotName` function later, in functions that will apply recommendation. This allows end users easier customization of our code.

pkg/automation/disks.go
Podkamienna: But then are unprintable characters ok?

minorczyka: I would just define array (or string) of alphanumeric characters and then I would sample from them.

pkg/automation/snapshotname_test.go
```diff
@@ -0,0 +1,65 @@
+/*
```
minorczyka, line 1: Convention is to name test files the same as the files they test, so in that case it would be disks_test.go

pkg/automation/disks.go
```diff
@@ -16,11 +16,72 @@ limitations under the License.
...
+)
+
+const maxDisknameLen = 20
+const maxZonenameLen = 20
+const maxSnapshotnameLen = 63
+
+const characters = "abcdefghijklmnopqrstuvwxyz" +
+	"ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+var generator = rand.New(rand.NewSource(time.Now().UnixNano()))
```
minorczyka, line 34: According to docs: "The default Source is safe for concurrent use by multiple goroutines, but Sources created by NewSource are not.". So you will have to either:
- protect it with mutex yourself
- use default Source (and seed it yourself)
- create generator in each function call

pkg/automation/snapshotname_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// Tests if the generated names are different
+func TestDifferentNames(t *testing.T) {
```
minorczyka, line 29: This isn't a really good test but testing nondeterministic behavior is hard in general.

If we have a bad luck (it depends on the seed we got) this test will fail. Flaky tests can be even worse then no tests at all. Unit tests should be deterministic and in this case you could achieve it by seeding source with some constant value.

pkg/automation/snapshotname_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Tests that the returned name is not to long if the disk name is long
+func TestLongDiskName(t *testing.T) {
+	testZone := ""
+	testDisk := strings.Repeat("a", 2*maxSnapshotnameLen)
+
+	result := randomSnapshotName(testZone, testDisk)
+
+	assert.LessOrEqual(t, len(result), maxSnapshotnameLen, fmt.Sprintf("The length of the returned snapshot name must be less than %d", maxSnapshotnameLen))
+}
```
minorczyka, line 65: I would also add one test that checks if `randomSnapshotName` returns what we expect: zone name, disk name, some timestamp and random characters at the end. You can just check that you can parse timestamp part (you don't have to compare it's value) and only check that some random characters are present.

PR #87, opened 2020-08-06
Frontend miscellaneous

s17k:
A few unrelated changes:
- **new:** added a global config file
- **update:** disabled fetching requests from the fake service when testing the Home component.
- **update:** default behaviour of fake service changed, so that no errors are present and so that negative costs (savings) are sent (so that we can see some green when visually testing the frontend).
- **update:** moved "selected" (list of selected recommendations, bound to the selected  rows in the table) to the coreTable Vuex module from the recommendations module.

### DIFF
--- a/frontend/src/views/Home.vue
+++ b/frontend/src/views/Home.vue
@@ -48,19 +48,9 @@ limitations under the License. -->
 import { Component, Vue } from "vue-property-decorator";
 import GroupingHint from "@/components/GroupingHint.vue";
 import CoreTable from "@/components/CoreTable.vue";
-import { ProjectConfig } from "./../config";
 
 @Component({
   components: { GroupingHint, CoreTable }
 })
-export default class Home extends Vue {
-  private mounted() {
-    /* In order for this fetch to work with the fake middleware service,
-     run: `go run cmd/fake-service/*.go` first from the root folder.
-    It might initially help to run it repeatedly until the installing 
-     errors disappear, make sure that Go is in the latest version too. */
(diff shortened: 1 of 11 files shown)


frontend/src/views/Home.vue
```diff
@@ -56,10 +57,10 @@ export default class Home extends Vue {
...
     /* In order for this fetch to work with the fake middleware service,
      run: `go run cmd/fake-service/*.go` first from the root folder.
-    It might sometimes help to run it repeatedly until the errors disappear,
-     make sure that Go is in the latest version too.
-    Finally, uncomment the following line: */
-    // this.$store.dispatch("recommendationsStore/fetchRecommendations");
+    It might initially help to run it repeatedly until the installing 
+     errors disappear, make sure that Go is in the latest version too. */
+    if (ProjectConfig.SEND_REQUESTS_FOR_RECOMMENDATIONS)
+      this.$store.dispatch("recommendationsStore/fetchRecommendations");
```
minorczyka, line 63: When I'm now thinking about it, we probably should put this dispatch somewhere else, e.g. to our main entry point. If we introduce another pages (e.g. with statistics and graphs that I mentioned today) then going back to home will trigger fetching recommendations every time. Placing it somewhere else will also more cleanly solve your problem with testing.

s17k: That is a very fair point!
I moved it to main.ts, so that it is dispatched even before the Vue root instance is created.
This is not a problem because the fetch only affects the store, and not Vue components.

PR #89, opened 2020-08-06
New data model, handling "apply" clicked.

s17k:
There are not as many changes as you might guess from the line count, as I have commented out the apply handler to keep this PR lean-ish.
### New data model
- **new:** extended RecommendationExtra class to store errors from applying recommendations 
- **update:** Because of that, RecommendationExtra objects for each recommendation are now kept in the store (previously they were computed on the go from the Recommendation objects, which didn't allow for any new data to be stored there). 
- **update:** all properties that are either inherited from the database object (previous interface name: Recommendation) or can be computed directly from that object are marked as `readonly` (but not recursively, because that would make testing annoying)
- **update:** renamed Recommendation -> RecommendationRaw 

### Error dialog UI, preparing for implementing the apply button handler
- **new:** Added a dialog that opens when "Show error" is clicked and shows the errors that are retrieved from the vuex store
- **new:** Vuex mutations in the RecommendationsStore module to update status and error headers/descriptions

### Fake-service issue
- **fix:** Slashes in the recommendation name are not parsed, so I have put the recommendation name as a query parameter instead.

### DIFF
--- a/frontend/src/store/recommendations.ts
+++ b/frontend/src/store/recommendations.ts
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 import {
-  Recommendation,
+  RecommendationRaw,
   RecommendationExtra,
   getRecommendationProject,
   getRecommendationType,
@@ -28,7 +28,7 @@ const REQUEST_DELAY = 100;
 const HTTP_OK_CODE = 200;
 
 export interface IRecommendationsStoreState {
-  recommendations: Recommendation[];
+  recommendations: RecommendationExtra[];
   selected: RecommendationExtra[];
 
   errorCode: number | undefined;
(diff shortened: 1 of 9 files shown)


s17k:
Sorry, this is not ready yet. I thought that it is, but I forgot about a few changes that I wanted to put here as well.
If you have reviewed some changes already, that is fine because they are mostly independent from the rest of what I am about to add here.

frontend/src/store/recommendations.ts
```diff
@@ -100,20 +130,151 @@ const actions: ActionTree<IRecommendationsStoreState, IRootStoreState> = {
...
     context.commit("endFetching");
-  } /* TODO: TO BE IMPLEMENTED,
-  async applySingleRecommendation(
+  }
+  /* [[FUTURE CHANGES, TO BE REVIEWED IN FURTHER PR-S TO KEEP THIS ONE SMALLER]]
+  // we need names, not references so that we can find them in the state fast
+  applyGivenRecommendations(context, selectedNames: string[]): void {
+    // if selected has duplicates
+    if (new Set(selectedNames).size !== selectedNames.length)
+      throw "Duplicates found among given recommendation names";
```
brzezinski, line 147: what if we will have multiple recommendations for one resource?

s17k: These are not resource names, but recommendation names (e.g. `recommendations/projects/323016592286/locations/us-central1-c/recommenders/google.compute.instance.IdleResourceRecommender/recommendations/6df88342-8116-441c-beb7-ab66d18a3078`). I could maybe change the name `selectedNames` to `selectedRecNames` to make this clearer.

(btw. I commented out these changes so that we can review them next time, I assume this is just something you noticed anyway and wanted to tell me about)

PR #92, opened 2020-08-11
Frontend apply logic

s17k:
- Handling 'apply' being clicked for one or more recommendations.
- Launching and running status watchers for recommendations (status watchers are launched at start for all 'CLAIMED' recommendations and whenever a recommendation is applied), converted from commented out code.
- Tests for both of these
- Also, fixed the root store factory to use module store factories instead of global constants. This caused some trouble for running multiple tests, as they shared some data.

### DIFF
--- a/frontend/tests/unit/apply.spec.ts
+++ b/frontend/tests/unit/apply.spec.ts
@@ -141,7 +141,7 @@ describe("watchStatus action", () => {
     await watchStatus(context, firstRec);
     expect(firstRec.statusCol).toBe(getInternalStatusMapping("CLAIMED"));
 
-    // end the status updates cycle
+    // dispatch watchStatus in the future
     expect(setTimeout as any).toBeCalledTimes(1);
   });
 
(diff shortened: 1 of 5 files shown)


frontend/tests/unit/apply.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+    expect(setTimeout as any).toBeCalledTimes(1);
+  });
+
+  test("-> succeeded", async () => {
+    fetchMock.mockResponseOnce(JSON.stringify({ status: "SUCCEEDED" }));
+    await watchStatus(context, firstRec);
+    expect((fetch as any).mock.calls[0][0].indexOf(firstRec.name)).not.toBe(-1);
+    expect(firstRec.statusCol).toBe(getInternalStatusMapping("SUCCEEDED"));
+
+    // dispatch watchStatus in the future
```
s17k, line 154: Lines 154 and 145 should be swapped.

PR #93, opened 2020-08-11
Apply

ViktoriiaKovalova:
add applying recommendation function


pkg/automation/apply.go
```diff
@@ -0,0 +1,108 @@
...
+const (
+	projectPath     = "projects"
+	zonePath        = "zones"
+	instancePath    = "instances"
+	diskPath        = "disks"
+	machineTypePath = "machineTypes"
+)
+
+// DoOperation does the action specified in the operation.
+func (s *googleService) DoOperation(operation *gcloudOperation) error {
```
minorczyka, line 41: I would like to have unit tests for `DoOperation` and `Apply` functions. We should have a test for every type of operation and for every type of recommendation we support.

pkg/automation/apply.go
```diff
@@ -0,0 +1,108 @@
...
+		default:
+			return errors.New(operationNotSupportedMessage)
+		}
+	case "replace":
+		if operation.ResourceType != "compute.googleapis.com/Instance" {
+			return errors.New(operationNotSupportedMessage)
+		}
+		switch operation.Path {
+		case "/machineType":
+			return s.replaceMachineType(operation)
```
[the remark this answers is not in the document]
Podkamienna: Actually it doesn't and it probably should (to change the machine type, it must be stopped first). Should I change the replaceMachineType function or call stopInstance before calling it and start instance after calling it?

pkg/automation/apply.go
Podkamienna: I don't have any good idea, in which way should the unit tests work. Can I ask for some hint?

Podkamienna: One of my ideas is adding a recommendation list to the test service and checking, if the changes in it are the ones I expect. Another idea is checking if the right functions are called with correct parameters.

pkg/automation/apply.go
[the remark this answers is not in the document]
minorczyka: If trying to change machine type of a running VM doesn't work (doesn't stop it, resize and restart) then we definitely should do it ourselves.

pkg/automation/apply.go
minorczyka: These unit tests should test if correct methods (with correct parameters) of `GoogleService` interface are called: `StopInstance`, `StartInstance`, `MarkRecommendationClaimed` etc. We introduced this interface so we can easily mock this calls.

You will most probably have problems with using mock because you declared these new functions as methods of `googleService`, you might have to change them to be normal functions.

pkg/automation/service.go
```diff
@@ -71,6 +71,12 @@ type GoogleService interface {
 
 	// starts the specified instance
 	StartInstance(project, zone, instance string) error
+
+	// tests if the machine type is correct
+	TestMachineType(project string, zone string, instance string, value interface{}, valueMatcher *gcloudValueMatcher) (bool, error)
+
+	// tests if the status is correct
+	TestStatus(project string, zone string, instance string, value interface{}, valueMatcher *gcloudValueMatcher) (bool, error)
```
minorczyka, line 79: I wouldn't add these functions to the interface. Right now, functions here correspond to similar calls from Compute API or Recommender API. While these two functions just have some of our logic and just call GetInstance method.

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,420 @@
...
+type calledFunction struct {
+	functionName string
+	arguments    []interface{}
+	results      []interface{}
+}
+
+type ApplyMockService struct {
+	GoogleService
+	calledFunctions []calledFunction
+}
```
minorczyka, line 24: testify library that we're already using has a mock package: https://github.com/stretchr/testify#mock-package

It might be shorter and cleaner than using your own solution. But I've never used it and I might be wrong so please just have a look and consider if we want to use it.

Podkamienna: I'm not sure - I mean, if we wanted to test if the right functions are called, I think it would be quite similar. But we could do the same thing one level lower - mock the calls to recommender API and compute API. But the issue with that is the methods of GoogleService, need to receive a server to perform further calls. It would be possible to mock that, but it could get quite long and complicated.

pkg/automation/service.go
Podkamienna: But then I need to mock the call to getInstance method. And that's an issue, because testing (TestMachineType and TestStatus) functions dereference the pointer to the returned machineInstance. It's quite difficult to tell what getInstance should return - what status and machine type should be set in the returned instance. So what I did is returning (true, nil) from the testing functions, when they receive (nil, nil) from the getInstance method. But I don't like this solution - it's not very elegant.

Podkamienna: But I also can set Value and ValueMatcher to nil in the recommendations in the tests, and I guess that this is good enough.

Podkamienna: Okay, in the end, I pass the result that GetInstance should return in the service struct.

pkg/automation/operation_handlers.go
```diff
@@ -0,0 +1,161 @@
...
+	path := operation.Resource
+
+	project, errProject := extractFromURL(path, projectParam)
+	zone, errZone := extractFromURL(path, zoneParam)
+	instance, errInstance := extractFromURL(path, instanceParam)
+	err := chooseNotNil(errProject, errZone, errInstance)
+	if err != nil {
+		return err
+	}
+
```
ViktoriiaKovalova, line 34: this code is repeating in testStatus, also GetInstance call in TestMachineType is similar to the code in TestStatus. I think we can make it a bit more readable & structured if we put that part into separate function.
for example:
```
 func TestInstanceFields(service GoogleService, operation ...) {
        path := operation.Resource
	project, errProject := extractFromURL(path, projectParam)
	zone, errZone := extractFromURL(path, zoneParam)
	instance, errInstance := extractFromURL(path, instanceParam)
	err := chooseNotNil(errProject, errZone, errInstance)
	if err != nil {
		return err
	}
        machineInstance, err := service.GetInstance(project, zone, instance)
	if err != nil {
		return false, err
	}
        var testField string
        switch operation.Path {
		case "/machineType":
			testField = machineInstance.MachineType
		case "/status":
			testField = machineInstance.Status
		default:
			return errors.New(operationNotSupportedMessage)
		}
       match, err := testMatching(testField, operation.Value, ...)
      if err != nil {
              //todo
      }
      if !match {
            return fmt.Errorf("field %s of instance is not as expected: equal to %s, not matching %v and %v", operation.Path, testField, operation.Value, operation.ValueMatcher) 
     }
     return nil
}
```

ViktoriiaKovalova: it's just a draft, i wrote it in comment on github, so there might be some errors, and some parts are missing:)

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1026 @@
...
+
+	for i := range functions {
+		result = append(result, calledFunction{functions[i], arguments[i], results[i]})
+	}
+
+	return result, nil
+}
+
+// Checks if two arrays of calledFunction type are equal
+func compareCalledFunctions(t *testing.T, expected, received []calledFunction) {
```
ViktoriiaKovalova, line 100: was assert.Equal(t, expected, received) not working?

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1026 @@
...
+		Action:       "test",
+		Path:         "/machineType",
+		Resource:     "//compute.googleapis.com/projects/rightsizer-test/zones/us-east1-b/instances/alicja-test",
+		ResourceType: "compute.googleapis.com/Instance",
+		ValueMatcher: &gcloudValueMatcher{MatchesPattern: ".*zones/us-east1-b/machineTypes/n1-standard-4"},
+	}
+
+	service := ApplyMockService{getInstanceResult: &compute.Instance{MachineType: "zones/us-east1-b/machineTypes/n1-standard-4"}}
+	err := DoOperation(&service, &operation)
+	assert.Nilf(t, err, "DoOperation shouldn't return an error")
```
ViktoriiaKovalova, line 129: You can use NoError method here: https://godoc.org/github.com/stretchr/testify/assert#NoError
(Doesn't really matter which one to use, but with NoError it's easier to understand that you want to check for no errors)

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1026 @@
...
+
+	expectedFunctions := []string{"GetInstance"}
+	expectedArguments := [][]interface{}{{"rightsizer-test", "us-east1-b", "alicja-test"}}
+	expectedResults := [][]interface{}{{&compute.Instance{MachineType: "zones/us-east1-b/machineTypes/n1-standard-4"}, nil}}
+
+	expected, _ := newCalledFunctions(expectedFunctions, expectedArguments, expectedResults)
+	compareCalledFunctions(t, expected, service.calledFunctions)
```
ViktoriiaKovalova, line 136: put this code in 'if' as in this example: https://godoc.org/github.com/stretchr/testify/assert#NoError
because we don't want to check it if checking for no errors already failed

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1026 @@
...
+
+	expectedFunctions := []string{"DeleteDisk"}
+	expectedArguments := [][]interface{}{{"rightsizer-test", "europe-west1-d", "vertical-scaling-krzysztofk-wordpress"}}
+	expectedResults := [][]interface{}{{nil}}
+
+	expected, _ := newCalledFunctions(expectedFunctions, expectedArguments, expectedResults)
+	compareCalledFunctions(t, expected, service.calledFunctions)
+}
+
+// Checks if receiveing an operation without necessary parameter
```
ViktoriiaKovalova, line 251: receiving*

pkg/automation/apply.go
```diff
@@ -0,0 +1,126 @@
...
+				errMark := service.MarkRecommendationFailed(recommendation.Name, recommendation.Etag)
+				if errMark != nil {
+					return errMark
+				}
+
+				return err
+			}
+		}
+	}
+	err = service.MarkRecommendationSucceeded(recommendation.Name, recommendation.Etag)
```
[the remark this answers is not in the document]
Podkamienna: Okay, thank you. I didn't realize that Etag will change.

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1026 @@
...
+		"MarkRecommendationClaimed",
+		"GetInstance",
+		"StopInstance",
+		"MarkRecommendationSucceeded",
+	}
+	expectedArguments := [][]interface{}{
+		{"projects/323016592286/locations/us-central1-a/recommenders/google.compute.instance.IdleResourceRecommender/recommendations/63378bdf-9ffe-4ea4-b8ee-04145f2a59c9", "\"9f58395697934a1a\""},
+		{"rightsizer-test", "us-central1-a", "vkovalova-instance-memory-1"},
+		{"rightsizer-test", "us-central1-a", "vkovalova-instance-memory-1"},
+		{"projects/323016592286/locations/us-central1-a/recommenders/google.compute.instance.IdleResourceRecommender/recommendations/63378bdf-9ffe-4ea4-b8ee-04145f2a59c9", "\"9f58395697934a1a\""},
```
ViktoriiaKovalova, line 318: maybe we can just use recommendation.Name, recommendation.Etag instead of copypasting these values? cause it's a bit hard to read and we can easily make some typo.
also, to make sure that these values are the same as when recommendation was declared, we can just make a copy  of recommendation and test with assert.Equal that recommendation stayed the same after Apply was called.

ViktoriiaKovalova: also, as i mentioned in previous comments, etag will change, so you can just add some symbol ('1', for example) to it in mock Mark methods, and check that in this case MarkRecommendationSucceeded was already called with recommendation.Etag + "1" or something like this.

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1026 @@
...
+		Name:      "projects/323016592286/locations/europe-west1-d/recommenders/google.compute.disk.IdleResourceRecommender/recommendations/1e32196d-fc39-4358-9c9b-cec17a85f4ea",
+		StateInfo: &gcloudStateInfo{State: "Active"},
+	}
+
+	service := ApplyMockService{calledFunctions: []calledFunction{}}
+	err := Apply(&service, &recommendation)
+	assert.Nilf(t, err, "Apply shouldn't return an error")
+
+	expectedFunctions := []string{
+		"MarkRecommendationClaimed",
```
ViktoriiaKovalova, line 366: it seems a bit not convenient for you to write this markrecommendationclaimed + succeeded every time. seems like   a bad design. i still think that we should separate actually applying recommendation in separate function. what do you think?

ViktoriiaKovalova: now as we'll change this etag updating you'll also have to update expected etag in second call to Mark in each of those tests, but if we'll make apply separate function, that we'll be called from ApplyAndClaim, then you'll only need to update tests testing ApplyAndClaim.

pkg/automation/apply_test.go
Podkamienna: Actually, I wanted to make the code simpler, by just assuming, that as this is a helper function, we will never receive an error. So then maybe I should just cause panic, if something is wrong, instead of returning an error? I mean, it will only return an error, if it receives arrays of different lengths.

pkg/automation/apply_test.go
Podkamienna: But I think that recommendation should change after apply?

pkg/automation/apply_test.go
Podkamienna: But then we would probably need to have another function, that calls 5/6 small functions and I think it would be easier to make a mistake or forget something. I can create another "constructor" for calledFunction type, which will insert the correct values at the beginning and end of calledFunctions array. I will change etag, by passing the recommendation to a function updateRecommendation, which will increment etag by one.

Podkamienna: Because, from what I understand, the recommendation, that is passed to apply, should be overwritten by the one returned from markClaimed, and later from markFailed/Succeeded?

pkg/automation/apply_test.go
ViktoriiaKovalova: why? you don't change it anywhere in apply function

Podkamienna: Yes, but I think I should. Because Etag of the recommendation changes.

pkg/automation/apply_test.go
ViktoriiaKovalova: i meant just having a function like:
```
func Apply(recommendation ...) err {
      rec, err := MarkClaimed(recommendation.Name, //etag)
      if err != nil {
           //todo
      }
      err = DoOperations(recommendation)
      if err != nil {
           _, _ := MarkFailed(//todo)
          return err
      }
      return nil
}
```

Podkamienna: Is it on purpose, that checking status and MarkSucceeded is not here?

ViktoriiaKovalova: I proposed it because it would make unit-testing easier, having small functions instead of one big function doing a lot of things.

Podkamienna: And you want to do that to be able to make the bigger part of the tests, to test DoOperations and not Apply? I think that it may be a good idea.

ViktoriiaKovalova: yes, that's way you don't have to add MarkClaimed + Suceeded to every test.
And for Apply we can just have 2 tests to check MarkClaimed + MarkFailed / MarkClaimed + MarkSuceeded is called in Apply as expected(if there is error/ no error).

pkg/automation/apply_test.go
ViktoriiaKovalova: ok, you are right, if you'll update recommendation that's fine

ViktoriiaKovalova: just try to avoid copypasting these lines

pkg/automation/apply_test.go
ViktoriiaKovalova: in my opinion you suggestion is okay, just leave some clear description of error in log.Fatal()

pkg/automation/apply_test.go
Podkamienna: Yes, that's actually a mistake. In other tests I used recommendation.Name.

pkg/automation/apply_test.go
Podkamienna: I didn't know, that Go was smart enough to be able to do it this way. That's quite cool.

pkg/automation/apply_test.go
ViktoriiaKovalova: > Is it on purpose, that checking status and MarkSucceeded is not here?

nope, i forgot it

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1097 @@
...
+		Action:       "test",
+		Path:         "/machineType",
+		Resource:     "//compute.googleapis.com/projects/rightsizer-test/zones/us-east1-b/instances/alicja-test",
+		ResourceType: "compute.googleapis.com/Instance",
+		ValueMatcher: &gcloudValueMatcher{MatchesPattern: ".*zones/us-east1-b/machineTypes/n1-standard-4"},
+	}
+
+	service := ApplyMockService{getInstanceResult: &compute.Instance{MachineType: "zones/us-east1-b/machineTypes/n1-standard-4"}}
+	err := DoOperation(&service, &operation)
+	assert.NoErrorf(t, err, "DoOperation shouldn't return an error")
```
ViktoriiaKovalova, line 126: `f` in the end of the method indicated that is can be used to format the text, like in Printf, so you don't really have to use it here. But it would be working fine, of course

pkg/automation/apply.go
```diff
@@ -0,0 +1,133 @@
...
+		return testInstanceField(service, operation)
+	case "replace":
+		if operation.ResourceType != "compute.googleapis.com/Instance" {
+			return errors.New(operationNotSupportedMessage)
+		}
+		switch operation.Path {
+		case "/machineType":
+			return replaceMachineType(service, operation)
+		case "/status":
+			if operation.Value != "TERMINATED" {
```
ViktoriiaKovalova, line 56: i would be nice to have constants for status of machine actually. it's used only once in this file, but you use it in tests frequently("ACTIVE"/"TERMINATED").

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1097 @@
...
+			},
+		},
+		Etag:      "\"40204a1000e5befe\"",
+		Name:      "projects/323016592286/locations/us-central1-a/recommenders/google.compute.instance.MachineTypeRecommender/recommendations/5df355d9-2f50-4567-a909-bcfcebcf7d66",
+		StateInfo: &gcloudStateInfo{State: "Active"},
+	}
+
+	service := ApplyMockService{}
+	err := DoOperations(&service, &recommendation)
+	assert.Error(t, err, operationNotSupportedMessage)
```
ViktoriiaKovalova, line 499: are you trying to test here that err will have message `operationNotSupportedMessage`?

pkg/automation/apply_test.go
```diff
@@ -0,0 +1,1097 @@
...
+							Action:       "replace",
+							Path:         "/machineType",
+							Resource:     "//compute.googleapis.com/projects/rightsizer-test/zones/us-central1-a/instances/sidharan-e2-with-stackdriver",
+							ResourceType: "compute.googleapis.com/Instance",
+							Value:        "zones/us-central1-a/machineTypes/e2-medium",
+						},
+					},
+				},
+			},
+		},
```
ViktoriiaKovalova, line 578: i'm not sure if you need to test whole DoOperations function if you want to see if some operation will fail. I think you can have tests for DoOperation on single operation for this purpose. And if you wanted to check that after one failed operation DoOperations won't proceed with others you can have a separate test as in TestFailedTest for it. I mean it would have been easier for you to do it this way, but it's not a problem

ViktoriiaKovalova (commented):
i have a few comments for the functions, but they look good.
tests are a bit hard to read because there are a lot of unnecessary code, like when you are testing that operation will fail, but actually testing whole recommendation with a few operations. It would be nice for every test to be focused on simple task, and if you want to test 2 things at once, maybe sometimes it's better to split it to 2 tests.

pkg/automation/apply_test.go
Podkamienna: Yes, I was trying to do that, and now I actually do that.

pkg/automation/apply.go
Podkamienna: Yes and the same goes for "replace" "remove" "test" and many other strings. I don't think, that this is a reason for defining a constant - I don't think that this would make the code easier to understand. If something was to change with the recommendation's content, then it would make it easier to adapt - but if something changed with recommendations, the whole app would probably stop working.

pkg/automation/apply_test.go
Podkamienna: I think you're right. I just didn't really know, how to write unite tests. Now I think I understand it better. Do you think, that it is worth it, to change them to be simpler and have more of them?

ViktoriiaKovalova: you can leave it as it is

pkg/automation/access.go
```diff
@@ -92,22 +92,17 @@ func (s *googleService) ListAPIRequirements(project string, apis []string) ([]Re
...
+	[]string{"compute.disks.createSnapshot", "compute.snapshots.create"},    // CreateSnapshot
+	[]string{"compute.disks.delete"},                                        // DeleteDisk
+	[]string{"compute.instances.get"},                                       // GetInstance
+	[]string{"recommender.computeDiskIdleResourceRecommendations.list"},     // ListRecommendations for google.compute.disk.IdleResourceRecommender
+	[]string{"recommender.computeInstanceIdleResourceRecommendations.list"}, // ListRecommendations for google.compute.instance.IdleResourceRecommender
+	[]string{"recommender.computeInstanceMachineTypeRecommendations.list"},  // ListRecommendations for google.compute.instance.MachineTypeRecommender
+	[]string{"compute.regions.list"},                                        // ListRegionsNames
+	[]string{"compute.zones.list"},                                          // ListZonesNames
+	[]string{"compute.instances.start"},                                     // StartInstance
+	[]string{"compute.instances.stop"},                                      // StopInstance
```
ViktoriiaKovalova: why have you deleted some of the permissions in this commit?

pkg/automation/apply.go
ViktoriiaKovalova: it's mostly needed to avoid some typos while writing code in this case.And generally it's a better practice than writing some repeatedly used strings in code, especially when they might be changed some time later, like error messages, for example. but it's just a minor comment, you can leave it if you want

pkg/automation/apply.go
```diff
@@ -0,0 +1,130 @@
...
+				return errors.New(operationNotSupportedMessage)
+			}
+
+			return stopInstance(service, operation)
+		default:
+			return errors.New(operationNotSupportedMessage)
```
ViktoriiaKovalova, line 62: could you please also remove default case here

pkg/automation/apply.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+	case "add":
+		switch operation.ResourceType {
+		case "compute.googleapis.com/Snapshot":
+			return addSnapshot(service, operation)
+		default:
+			return errors.New(operationNotSupportedMessage)
+		}
```
ViktoriiaKovalova, line 70: and here?

ViktoriiaKovalova:
please make sure every file have license header (now some don't)

pkg/automation/access.go
Podkamienna: I don't know, I copied them from another commit and probably some of them were hidden, because they were not changed, and I didn't notice that.

ViktoriiaKovalova:
Hi! It's a national holiday today, so I'm OOO, don't wait for me on our meeting

minorczyka (approved):

PR #99, opened 2020-08-13
Frontend footer creation

Podkamienna:
recomator1
recomator2


frontend/src/components/FooterSummary.vue
```diff
@@ -0,0 +1,175 @@
...
+    for (const recommendation of selectedRecommendations) {
+      if (recommendation.costCol < 0) {
+        result -= recommendation.costCol; // Minus to have absolute value
+      }
+    }
+
+    return result;
+  }
+
+  get spendingsFromSelected(): number {
```
brzezinski, line 99: is it better to use abs() and leave only one function instead of two? They are doing almost same job

Podkamienna: I am not sure if that's a good idea. The reason, is that I would need to pass some predicate as an argument in order to decide, when to add a value to the result. And I can't do that with get - they can't have arguments. I could use a method to do that, but then the function would be computed each time. And when using get, it would be computed, only when some of its dependencies changed. On the other hand, the list of selected recommendations may change very often and then using computed properties instead of methods may not be that beneficial.

frontend/src/components/FooterSummary.vue
```diff
@@ -0,0 +1,175 @@
...
+
+  get savingsPart(): string {
+    const savings = this.savingsFromSelected;
+    if (savings === 0) {
+      return "";
+    }
+
+    return ` Save ${savings.toFixed(
+      2
+    )}$ each week by not using unnecessary resources.`;
```
s17k, line 142: Maybe "by right-sizing your instances" would sound better ?
Or "by using less resources" ?

frontend/src/components/FooterSummary.vue
```diff
@@ -0,0 +1,175 @@
...
+<template>
+  <div>
+    <v-footer
+      v-if="selectedRowsCount > 0"
+      v-bind:fixed="true"
+      color="rgba(255, 210, 20, 1)"
+      class="pa-6"
+    >
+      <v-row class="px-5">
+        <v-spacer></v-spacer>
```
s17k, line 23: you should be able use `<elem/>` instead of `<elem></elem>` to make it more concise

frontend/src/components/FooterSummary.vue
```diff
@@ -0,0 +1,175 @@
+<!-- Copyright 2020 Google LLC
```
s17k, line 1: Why not `Footer.vue` ? There is much more than a summary implemented here (like the 'are you sure' dialog and 'apply' button).

frontend/src/components/FooterSummary.vue
```diff
@@ -0,0 +1,175 @@
...
+  get selectedRows(): RecommendationExtra[] {
+    return (this.$store.state as IRootStoreState).coreTableStore!.selected;
+  }
+
+  get selectedRowsCount(): number {
+    return this.selectedRows.length;
+  }
+
+  get savingsFromSelected(): number {
+    const selectedRecommendations = this.selectedRows;
```
s17k, line 87: Why not just use `this.selectedRows` here and in other places ?

frontend/src/components/FooterSummary.vue
```diff
@@ -0,0 +1,175 @@
...
+      color="rgba(255, 210, 20, 1)"
+      class="pa-6"
+    >
+      <v-row class="px-5">
+        <v-spacer></v-spacer>
+
+        <div class="font-weight-black">
+          {{ footerMessage }}
+        </div>
+        <v-spacer></v-spacer>
```
s17k, line 28: You can use `<v-col>`s instead of `<v-spacer>`s for centring.
Vuetify sample pages do use `<v-row>`'s without `<v-col>`s inside of them sometimes but they also say that `v-row` [is just a wrapper](https://vuetifyjs.com/en/components/grids/) for `v-col`.

frontend/src/components/FooterSummary.vue
s17k: You could use `filter`, `reduce` so that they are much more concise.

frontend/src/components/FooterSummary.vue
brzezinski: +1 for "by using less resources"

Podkamienna: But what is wrong with "using unnecessary resources"? Is it not in English? Or is the double negation wrong? I have some doubts about writing "by using less resources", because it does not contain information about those resources not being needed for anything.

s17k: 'Resources' is actually quite general and encompasses all the infrastructure "rented" from Google Cloud (like disks/vms/network utilities).

Podkamienna: Is that an issue? The client will be able to see the detailed description of each recommendation.

s17k: I was replying to your doubts about writing "using less resources", it is not an issue in and of itself.

brzezinski: as I told you during our daily call giving double negation "not using" "unnecessary" may be harder to understand by user. That's why I agreed with proposal "by using less resources"

PR #100, opened 2020-08-13
add listing all recommendations & progress tracking

ViktoriiaKovalova:


pkg/automation/utility.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// GetProgress returns the fraction of work done
+func (t *Task) GetProgress() (int32, int32) {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+	if t.taskDone {
+		return 1, 1
+	}
+	answer := 0.0
```
minorczyka, line 86: It was really not obvious what's happening here when I've first seen it. I think that naming this variable "ratio" instead of "anwer" would help and changing name of a `floatToInts` function would also be nice, now it's not really descriptive. Describing in comment meaning of these two ints returned by this function would also be good.

pkg/automation/listrecommendations.go
```diff
@@ -91,7 +91,12 @@ func (s *googleService) ListRegionsNames(project string) ([]string, error) {
 // Requires the recommender.*.list IAM permission for the specified recommender.
 // numConcurrentCalls specifies the maximum number of concurrent calls to ListRecommendations method,
 // non-positive values are ignored, instead the default value is used.
-func ListRecommendations(service GoogleService, project, recommenderID string, numConcurrentCalls int) ([]*gcloudRecommendation, error) {
+func ListRecommendations(service GoogleService, project, recommenderID string, numConcurrentCalls int, tasks ...*Task) ([]*gcloudRecommendation, error) {
+	var task *Task
+	if len(tasks) != 0 {
+		task = tasks[0]
+	}
```
minorczyka, line 98: I don't like using this variadic arguments for optional parameters, it makes interface of these function not obvious. I would just add normal argument `task *Task`.

pkg/automation/listrecommendations.go
```diff
@@ -147,5 +155,108 @@ func ListRecommendations(service GoogleService, project, recommenderID string, n
 	if err != nil {
 		return nil, err
 	}
+	defer task.SetAllDone()
```
minorczyka, line 158: You don't need defer here.

pkg/automation/utility.go
```diff
@@ -0,0 +1,99 @@
...
+*/
+
+package automation
+
+import (
+	"sync"
+)
+
+// Task is the structure that helps get the percentage of work done
+type Task struct {
```
minorczyka, line 24: I like design of this Task struct. If we will need it, we can also easily implement `Cancel` method that would allow easily interrupting task and all its subtasks.

pkg/automation/access.go
```diff
@@ -46,41 +48,41 @@ type Requirement struct {
 // services.get permission is required for this method.
 // First requirement in returned list will be related to Service Usage API.
 // If it is not enabled or user doesn't have services.get permission other APIs won't be checked.
-func (s *googleService) ListAPIRequirements(project string, apis []string) ([]Requirement, error) {
+func (s *googleService) ListAPIRequirements(project string, apis []string) ([]*Requirement, error) {
```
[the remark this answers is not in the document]
ViktoriiaKovalova: I noticed that in go client library pointers are used everywhere in this structs, and I thought that there may be some some problems with copying to much, so probably it's a good practice to use pointers but I wasn't really sure.

pkg/automation/utility.go
```diff
@@ -0,0 +1,99 @@
...
+	subtasks     []Task
+	subtasksDone int
+	taskDone     bool
+	mutex        sync.Mutex
+}
+
+// AddSubtasks is used to add subtasks to the task
+func (t *Task) AddSubtasks(num int) {
```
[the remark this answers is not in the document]
ViktoriiaKovalova: yes, i'll probably change the name

pkg/automation/utility.go
ViktoriiaKovalova: yes, I agree, I'll try to make it clear:)

pkg/automation/access.go
```diff
@@ -92,38 +94,90 @@ func (s *googleService) ListAPIRequirements(project string, apis []string) ([]Re
...
+// If all statuses are equal to RequirementCompleted, user has all required permissions.
+func ListProjectRequirements(s GoogleService, project string) ([]*Requirement, error) {
+	requirements, err := s.ListAPIRequirements(project, requiredAPIs)
+	if err != nil {
+		return nil, err
+	}
+	ok := true
+	for _, req := range requirements {
+		if req.Status == RequirementFailed {
+			ok = false
```
Podkamienna, line 155: You also could have a return inside the "if req.Status == RequirementFailed" and then the ok variable would not be needed. I'm not sure if that's better stylistically, but it makes the function shorter and in my opinion more intuitive.

pkg/automation/utility.go
```diff
@@ -0,0 +1,97 @@
...
+*/
+
+package automation
+
+import (
+	"sync"
+)
+
+// Task is the structure that helps get the percentage of work done
+type Task struct {
```
Podkamienna, line 24: Maybe this structure and its methods could be in a different file?

pkg/automation/utility.go
```diff
@@ -0,0 +1,97 @@
...
+
+// IncrementDone increments subtasksDone
+func (t *Task) IncrementDone() {
+	t.mutex.Lock()
+	t.subtasksDone++
+	t.mutex.Unlock()
+}
+
+// GetNextSubtask returns the pointer to next not done subtask
+func (t *Task) GetNextSubtask() *Task {
```
Podkamienna, line 46: I am not sure about this, but maybe you should have an additional parameter of the task structure, showing which task should be next? I think that right now it is possible to have one task done twice.

pkg/automation/utility_test.go
```diff
@@ -0,0 +1,75 @@
+package automation
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
```
Podkamienna, line 8: I think there should be a test testing behavior of tasks with more than one thread.

pkg/automation/access.go
```diff
@@ -29,8 +29,10 @@ import (
...
 const (
-	requirementFailed    = "FAILED"
-	requirementCompleted = "COMPLETED"
+	// RequirementFailed corresponds to the failed status of requirement
+	RequirementFailed = "FAILED"
+	// RequirementCompleted corresponds to the completed status of requirement
+	RequirementCompleted = "COMPLETED"
 )
 
 // Requirement contains information about the required permission or api.
```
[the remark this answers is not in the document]
ViktoriiaKovalova: mmm?

pkg/automation/access.go
ViktoriiaKovalova: thanks, i'll change it

pkg/automation/utility.go
ViktoriiaKovalova: yes, i also thought about splitting it. i'll create a new file

pkg/automation/utility.go
ViktoriiaKovalova: if someone will forget to call IncrementDone() the GetNextSubtask() will still return the same task as before. Is this the issue you were talking about?

ViktoriiaKovalova: you're right, i'll fix this

ViktoriiaKovalova: actually, it won't cause a problem the way we are using this structure. 
we only need it to be multithread in a way that we can look at the progress from different threads.
i'll probably add a comment there

ViktoriiaKovalova: and add additional parameter

pkg/automation/listrecommendations.go
```diff
@@ -87,65 +87,171 @@ func (s *googleService) ListRegionsNames(project string) ([]string, error) {
...
+
+	results := make(chan recommendationsResult, numberOfQueries)
+	queries := make(chan query, numberOfQueries)
 
 	for i := 0; i < numWorkers; i++ {
 		go func() {
-			for location := range locationsJobs {
-				recs, err := service.ListRecommendations(project, location, recommenderID)
-				results <- result{recs, err}
+			for q := range queries {
```
brzezinski, line 169: I would recommend to use longer variable names like "query" instead of q

pkg/automation/access.go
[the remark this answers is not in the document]
Podkamienna: In the comment that is lower, if think there is a typo. But I couldn't comment on that line. So it's one or two lines below.

pkg/automation/listrecommendations.go
```diff
@@ -87,65 +87,171 @@ func (s *googleService) ListRegionsNames(project string) ([]string, error) {
...
+// Otherwise, projects requirements, including failed ones, are added to `failedProjects` to help show warnings to the user.
+// task structure tracks how many subtasks have been done already.
+func ListAllProjectsRecommendations(service GoogleService, numConcurrentCalls int, task *Task) (*ListResult, error) {
+	projects, err := service.ListProjects()
 	if err != nil {
 		return nil, err
 	}
-	return recommendations, nil
+
+	task.SetNumberOfSubtasks(2)
```
brzezinski, line 239: why is it 2?

brzezinski (approved):
LGTM

pkg/automation/listrecommendations.go
ViktoriiaKovalova: because there are to calls inside this function: ListRequirements and listRecommendationsIfRequirementsCompleted

brzezinski: maybe it's good to add some comment

ViktoriiaKovalova: Sure

PR #110, opened 2020-08-17
Frontend cost filter

s17k:
Closes: #110 
**After**:
Screenshot from 2020-08-17 17-49-39 (1)

### DIFF
--- a/frontend/src/store/core_table_filter_utils.ts
+++ b/frontend/src/store/core_table_filter_utils.ts
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 import { RecommendationExtra } from "./model";
-import { ICoreTableStoreState } from "./core_table";
+import { ICoreTableStoreState, costCategoriesNames } from "./core_table";
 
 // The primary motivation for having these in a separate file is so that
 //  they can be mocked by Jest:
@@ -54,6 +54,20 @@ export function statusFilterAccepted(
   return (
     tableState.statusesSelected.length === 0 ||
     tableState.statusesSelected.includes(recExtra.statusCol)
+  );
+}
+
+export function costFilterAccepted(
+  tableState: ICoreTableStoreState,
+  recExtra: RecommendationExtra
+) {
+  const catgsSelected = tableState.costCategoriesSelected;
+  const costsAlias = costCategoriesNames.costs;
+  const gainsAlias = costCategoriesNames.gains;
+  return (
+    catgsSelected.length == 0 ||
+    (recExtra.costCol >= 0 && catgsSelected.includes(costsAlias)) ||
+    (recExtra.costCol <= 0 && catgsSelected.includes(gainsAlias))
   );
 }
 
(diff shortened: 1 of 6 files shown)


frontend/tests/unit/aggregate_filter.spec.ts
```diff
@@ -40,12 +41,13 @@ describe("Filtering aggregate (resource name, type, status...)", () => {
     ) as boolean;
 
   test("project filter", () => {
-    let isOn = new Array(5).fill(true);
+    let isOn = new Array(6).fill(true);
```
Podkamienna, line 44: Maybe you can create a constant for the number of filters, so the code is easier to understand?

frontend/tests/unit/aggregate_filter.spec.ts
```diff
@@ -19,7 +19,8 @@ import {
   statusFilterAccepted,
   projectFilterAccepted,
   typeFilterAccepted,
-  resourceFilterAccepted
+  resourceFilterAccepted,
+  costFilterAccepted
```
Podkamienna, line 23: Why do all the filters have the suffix "Accepted"?

frontend/src/store/core_table_filter_utils.ts
```diff
@@ -57,6 +57,20 @@ export function statusFilterAccepted(
...
 
+export function costFilterAccepted(
+  tableState: ICoreTableStoreState,
+  recExtra: RecommendationExtra
+) {
+  const catgsSelected = tableState.costCategoriesSelected;
+  const costsAlias = costCategoriesNames.costs;
+  const gainsAlias = costCategoriesNames.gains;
+  return (
+    catgsSelected.length == 0 ||
```
Podkamienna, line 68: Is it really worth it, to define those constants? I think they make this function more difficult to understand. Moreover, for me it was not that easy to connect catgs to the word categories.

frontend/tests/unit/aggregate_filter.spec.ts
s17k: Because if you don't do that, it is hard to figure out what `resourceFilter(..., recommendation)` should return (should `true` be returned for the items that are filtered out or for the items that should be included in the results).

frontend/tests/unit/aggregate_filter.spec.ts
s17k: I have already made an effort not to make the `mockImplementation` calls too abstract (avoided the for loop), so I think it should really be quite easy to see what I am doing with `isOn`.

frontend/src/store/core_table_filter_utils.ts
s17k: Interesting, I will take a look and try to correct that.

frontend/tests/unit/aggregate_filter.spec.ts
brzezinski: magic numbers is something we should avoid

s17k: The magic bubble quickly bursts the moment you read the next few lines in my opinion 😃...

PR #111, opened 2020-08-18
Frontend get short name

Podkamienna:


frontend/src/store/model.ts
```diff
@@ -62,10 +62,60 @@ export interface OperationsList {
...
+  name: string;
+  source_disk: string;
+  storage_locations: string[];
+}
+
 export interface Operation {
   resource: string;
   resourceType: string;
   path: string;
+  value?: string | ValueAddSnapshot;
```
s17k, line 75: You could add a comment that this is part of a union field that you only support one type of (`value`, not `valueMatcher`) . The other type doesn't look like it is currently used in any recommendations, so this should be fine but I would make this decision more visible.

frontend/src/store/model.ts
```diff
@@ -75,12 +125,45 @@ export function getRecommendationResource(
...
 }
 
+// -> "compute.googleapis.com/Snapshot"
+export function getRecommendationResourceType(
+  recommendation: RecommendationRaw
+): string {
+  return recommendation.content.operationGroups[0].operations[0].resourceType;
+}
+
+export function getRecommendationValue(
```
s17k, line 135: This is not `Recommendation`, but `Operation` value according to the docs. Probably best not to name it this way then, I initially thought that it is a property of `Recommendation`.

frontend/src/store/model.ts
Podkamienna: Actually valueMatcher is used in the recommendation, that tests if the machine type is correct. Do you think, that then it also should be added?

s17k: I understand you mean the operation that tests the machine type. Is that an operation that you have seen in any recommenders that we support ?

frontend/src/store/model.ts
```diff
@@ -75,12 +125,45 @@ export function getRecommendationResource(
...
-  return extractFromResource("instances", resource);
+  const resourceType = getRecommendationResourceType(recommendation);
+
+  switch (resourceType) {
+    case "compute.googleapis.com/Instance": {
+      const resource = getRecommendationResource(recommendation);
+      return extractFromResource("instances", resource);
+    }
+    case "compute.googleapis.com/Disk": {
+      const resource = getRecommendationResource(recommendation);
```
s17k, line 153: It seems like the docs ([link](https://cloud.google.com/compute/docs/disks/viewing-and-applying-idle-pd-recommendations)) have an example of a recommendation with `resourceType` equal to `compute.googleapis.com/Disk` but with no `resource` field in the first operation. Maybe they are outdated, as the sample snapshot recommendations in the fake-service do have it, or we used an old version of the API to get these. Anyway, I think it is safer to just use the `resource` field from the second operation.

frontend/src/store/model.ts
```diff
@@ -62,10 +62,60 @@ export interface OperationsList {
   [index: number]: Operation;
 }
 
+export interface ValueAddSnapshot {
```
s17k, line 65: Naming it `OperationValueAddSnapshot` would make it less ambiguous (I know that `value` is an actual name of the field used in the docs, but barely anyone remembers that and it wasn't too easy to guess that).

frontend/src/store/model.ts
Podkamienna: Did you manage to find something in the docs about this operation being always followed by another one?

frontend/src/store/model.ts
Podkamienna: Yes, it's the first one in recommendations.go file. If by recommender you mean recommenderSubtype field, then it is present in CHANGE_MACHINE_TYPE.

frontend/src/store/model.ts
Podkamienna: Then we need to change also the functions "getRecommendationResource" and "getRecommendationResourceType". Actually, even though it is not a property of recommendation, but of operation I am not sure if the name is wrong. We use the value that is returned to represent the recommendation. Otherwise we would get the value that represents an operation, and it would be quite unclear, why do we use it to extract information about a recommendation.

frontend/src/store/model.ts
Podkamienna: I guess I will just check all the recommendations, and condition on operation's action instead of resource type.

Podkamienna: And throw an exception, when received names don't agree.

frontend/src/store/model.ts
s17k: My main point is that 'value' is super ambiguous and it is hard to guess that someone actually designed a field with such a name.

frontend/src/store/model.ts
s17k: > Did you manage to find something in the docs about this operation being always followed by another one?

No, but I don't think the basic structure of operations varies among recommendations from the same type.

frontend/src/store/model.ts
```diff
@@ -118,59 +121,70 @@ function isValueAddSnapshot(
...
-    }
-    case "compute.googleapis.com/Disk": {
-      const resource = getRecommendationResource(recommendation);
-      return extractFromResource("disks", resource);
-    }
-    case "compute.googleapis.com/Snapshot": {
-      const value = getRecommendationValue(recommendation);
-      if (isValueAddSnapshot(value)) {
+  switch (action) {
+    case "add": {
```
s17k, line 158: As long as this works, it is an acceptable solution.
But if it stops because, for example, we find out that not all recommendations that have a first 'add' operation are snapshot recommendations (as it is assumed now), I would definitely argue that we could do something simpler like this:
- if operations[0].value.source_disk exists, return the disk name
- if not, just return whatever is after the last slash in operations[0].resource

s17k (approved):
Looks good for now. I left a comment about potential changes in the future.

PR #113, opened 2020-08-21
Frontend cleanup - phase one

s17k:
Primarily renaming, more comments and of improved quality
Also, 
- fix: `<v-row>`s should be in `<v-containers>` according to Vuetify docs. Mobile view has two rows now, all larger ones have one to display the summary and 'apply' button. [related changes](https://github.com/googleinterns/recomator/pull/113/files#diff-fc0c24dc6322dfd54b09c2409f44088d)
- fix: We shouldn't have been recalculating column values from stored API instances, as they are already stored (they didn't use to, that is why the recalculation was implemented at some point). [commit](https://github.com/googleinterns/recomator/pull/113/commits/70ad2d319771a7ad54abed8a6f4f0350bbec6b71) Closes: #119


s17k:
Just to clarify, the file `model.ts` was split into `recommendation_raw.ts` and `recommendation_extra.ts`. The related added lines are not actually new.

frontend/src/components/Footer.vue
```diff
@@ -53,19 +55,20 @@ limitations under the License. -->
...
-            Proceed
+            <v-icon>mdi-check</v-icon>
+            Yes
           </v-btn>
 
-          <v-btn color="red white--text" v-on:click="dialog = false">
-            Resign
+          <v-btn color="primary white--text" v-on:click="dialog = false">
+            <v-icon>mdi-window-close</v-icon>
+            Cancel
```
Podkamienna, line 71: Why do you think that blue is better for the cancel button?

frontend/src/components/Footer.vue
```diff
@@ -53,19 +55,20 @@ limitations under the License. -->
...
-
           <v-btn
             color="green white--text"
             v-on:click="
               applySelectedRecommendations();
               dialog = false;
             "
           >
-            Proceed
+            <v-icon>mdi-check</v-icon>
```
Podkamienna, line 65: Is that possible for the icons to be a bit higher, so that the text is in the middle of them? I know, that's just a minor detail, but it doesn't look right in my opinion.

frontend/src/components/CoreTable.vue
```diff
@@ -78,7 +80,7 @@ import { RecommendationExtra } from "../store/model";
   }
 })
 export default class CoreTable extends Vue {
-  // headers ending with "Col" have values that are bound to corresponding properties
+  // headers ending with "Col" have values that are bound to corresponding columns
```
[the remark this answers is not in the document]
s17k: I am not a fan of consistency for the sake of consistency.
It is sometimes nice to have a certain decorum, but usually it also helps you differentiate between let's say class names and interface names. That is not the case here.

frontend/src/components/CoreTable.vue
```diff
@@ -105,6 +107,7 @@ export default class CoreTable extends Vue {
     { text: "", value: "statusCol", sortable: false }
   ];
 
+  // Sync selected with the store
```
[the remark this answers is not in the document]
s17k: Yes, https://dictionary.cambridge.org/dictionary/english/sync

frontend/src/components/Footer.vue
s17k: Because a red button associates with something really important (if you google 'nuclear button', all buttons are red). Blue is the colour that we use for nearly all other buttons (`primary`).

frontend/src/components/CoreTable.vue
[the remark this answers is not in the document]
s17k: It is extremely common in programming slang.

frontend/src/components/Footer.vue
s17k: Nicely spotted, should be doable.

frontend/src/components/Footer.vue
Podkamienna: Ok, that may be a good way to think about it.

Podkamienna (approved):

frontend/src/components/Footer.vue
brzezinski: agree

brzezinski (changes requested):
please review comments and update

PR #115, opened 2020-08-21
Waiting for googleService requests to finish

s17k:
- Added waiting for non-get requests to finish before returning.

~~I will add tests once this is pre-approved, as agreed.~~
Maybe it's not too useful to actually make tests for these, as they would need to be very tightly coupled with the implementation.

As of now, these have been tested and work (verified on Pantheon individually)
- `CreateSnapshot` 
- `StopInstance`
- `DeleteDisk`
- `ChangeMachineType`

### DIFF
--- a/pkg/automation/service.go
+++ b/pkg/automation/service.go
@@ -106,13 +106,13 @@ func NewGoogleService(ctx context.Context, conf *oauth2.Config, tok *oauth2.Toke
 	}, nil
 }
 
-// for anonymous functions passed to AwaitUntilCompletion
+// for anonymous functions passed to AwaitCompletion
 type operationGenerator func() (*compute.Operation, error)
 
-// AwaitUntilCompletion takes a function that needs to be called repeatedly
+// AwaitCompletion takes a function that needs to be called repeatedly
 //  to check if a process (some Google Service request) has finished.
 // Such a function is usually constructed by wrapping a .Do() call
-func AwaitUntilCompletion(gen operationGenerator) error {
+func AwaitCompletion(gen operationGenerator) error {
 	for {
 		oper, err := gen()
 		if err != nil {
--- a/pkg/automation/service.go
+++ b/pkg/automation/service.go
@@ -18,6 +18,7 @@ package automation
 
 import (
 	"context"
+	"time"
 
 	"golang.org/x/oauth2"
 	"google.golang.org/api/cloudresourcemanager/v1"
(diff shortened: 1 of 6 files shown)


pkg/automation/service.go
```diff
@@ -104,3 +105,22 @@ func NewGoogleService(ctx context.Context, conf *oauth2.Config, tok *oauth2.Toke
 		serviceUsageService:    serviceUsageService,
 	}, nil
 }
+
+// for anonymous functions passed to AwaitUntilCompletion
+type operationGenerator func() (*compute.Operation, error)
+
+// AwaitUntilCompletion takes a function that needs to be called repeatedly
+//  to check if a process (some Google Service request) has finished.
```
Podkamienna, line 113: Is it on purpose, that this line has an additional space at the beginning?

ViktoriiaKovalova (commented):
LGTM, I think tests for AwaitUntilCompletion with mock operation generator will be useful

pkg/automation/service.go
s17k: Yes. I like to do that to show that this is a continuation of the previous sentence.

brzezinski (approved):
LGTM

pkg/automation/service.go
minorczyka: I've never seen such convention. Usually if you really want to separate comments into more parts for better readability, you just leave blank line.

pkg/automation/service_test.go
```diff
@@ -0,0 +1,67 @@
...
+	assert.Equal(t, longOperationGenCounter, 3)
+
+	//ending in failure
+	longOperationGenCounter = 0
+	longOperationGenShouldSucceed = false
+	err = AwaitCompletion(longOperationGen)
+	assert.EqualError(t, err, "Oh no")
+	assert.Equal(t, longOperationGenCounter, 3)
+}
+
```
minorczyka, line 46: nit: It's better to split this function into three separate. Then you don't need these comments, you can communicate what's the purpose of test with it's name. Based on name it's also easier to tell which tests failed. Moreover if first assertion fails, rest of the test won't be run.

pkg/automation/service_test.go
```diff
@@ -0,0 +1,67 @@
...
+func fastFailingOperationGen() (*compute.Operation, error) {
+	return nil, errors.New("Oh no")
+}
+
+var longOperationGenCounter int
+var longOperationGenShouldSucceed bool
+
+func longOperationGen() (*compute.Operation, error) {
```
minorczyka, line 54: It would be better to create some struct that encapsulates these variables instead of using globals. Then you can turn `longOperationGen` into method of that struct.

minorczyka (changes requested):

pkg/automation/service_test.go
s17k: Will I be able to pass it as a function with no receivers and so that the context is maintained later though ?
`longOperationGen` is passed to `AwaitCompletion` which requires `func() (*compute.Operation, error)`

minorczyka: You will have to use lambda for it.

s17k: Yeah, I thought that it is better not do that as it felt like its gonna look 'smarter' (more abstraction, less globals), but actually harder to understand. It is going to be a generator of Operation generators after all.

s17k: You will be able to judge for yourself. I have just removed the globals.

minorczyka: Now it's fine.

minorczyka (approved):

PR #117, opened 2020-08-24
Frontend applied

Podkamienna:


frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +117,78 @@ export default class CoreTable extends Vue {
...
+      for (const row of this.selectableRows) {
+        if (this.isSelected(row)) {
+          this.selectedRows = this.selectedRows.filter(item => item !== row);
+        }
+      }
+    }
+  }
+
+  areAllSelected(): boolean {
+    for (const row of this.selectableRows) {
```
s17k, line 167: You can make this a one-liner with [every()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)

frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +117,78 @@ export default class CoreTable extends Vue {
...
+  }
+
+  isSelected(item: RecommendationExtra): boolean {
+    return this.selectedRows.includes(item);
+  }
+
+  select(item: RecommendationExtra, value: boolean) {
+    if (value) {
+      this.selectedRows = this.selectedRows.filter(elt => elt != item);
+    } else {
```
s17k, line 187: This has a potential to be quite slow, really and it's not obvious why is it done here.
Why not just add a one-line mutation to the store (let's say `addSelectedRow` and use it)

s17k: You could also use it in `selectAll`

frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +117,78 @@ export default class CoreTable extends Vue {
     { text: "", value: "statusCol", sortable: false }
   ];
 
+  // Array that keeps all recommendations that are selected
```
s17k, line 120: If it keeps all recommendations that are selected, why is it called `selectableRows` and not `selectedRows` ?

Podkamienna: No, it keeps all the recommendations from the current view, that can be selected. And it is needed for checking if all the recommendations are selected. I couldn't come up with a better way, but I am open for suggestions.

Podkamienna: I forgot to change the comment

frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +117,78 @@ export default class CoreTable extends Vue {
...
+  }
+
+  isActiveRecommendation(item: RecommendationExtra) {
+    return item.statusCol === getInternalStatusMapping("ACTIVE");
+  }
+
+  isSelected(item: RecommendationExtra): boolean {
+    return this.selectedRows.includes(item);
+  }
+
```
s17k, line 183: I would avoid calling it `value`, as it hardly explains what it actually is (hard to guess without looking at the code, my initial guess was actually wrong).
Why not call it `delete` or `toDelete` or `remove` or something like that ?

frontend/src/components/CoreTable.vue
Podkamienna: I agree that it's better to add the mutation to the store.

frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +117,78 @@ export default class CoreTable extends Vue {
...
   get selectedRows(): RecommendationExtra[] {
     return (this.$store.state as IRootStoreState).coreTableStore!.selected;
   }
 
   set selectedRows(selected: RecommendationExtra[]) {
     this.$store.commit("coreTableStore/setSelected", selected);
   }
+
+  pagination(page: any) {
```
s17k, line 139: It is quite hard to guess what this returns without remembering how the `pagination` slot works.

frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +117,78 @@ export default class CoreTable extends Vue {
     { text: "", value: "statusCol", sortable: false }
   ];
 
+  // Array that keeps all recommendations that are selected
+  // The array in the store keeps only the recommendations that
+  // are selected and are applicable
+  firstRow = 0;
+  lastRow = Infinity;
+  selectableRows = new Array<RecommendationExtra>();
```
s17k, line 125: It's not a rule, but it would be best to keep as much as possible in Vuex stores so that application state can be more easily tracked. Also, importing individual function from `.vue` files is a pain (at least I couldn't find a neat way), so probably best for all the logic that needs to be tested to be in stores.

frontend/src/components/CoreTable.vue
```diff
@@ -18,12 +18,24 @@ limitations under the License. -->
...
     :single-select="false"
-    v-model="selectedRows"
     show-select
+    v-model="selectedRows"
+    v-on:pagination="pagination"
     item-key="name"
     :footer-props="{ itemsPerPageOptions: [10, 100, -1] }"
   >
     <!-- ^customFilter prop is not used, because its implementation executes it for each property -->
+    <template v-slot:header.data-table-select="{}">
```
s17k, line 28: This overwrites the default behaviour which shows a minus if there is a mix of selected and not selected. I checked and Gmail has the exact same behaviour. I would trust their UX team.
You can just show an icon or button with an icon instead if that is the case.

frontend/src/components/CoreTable.vue
s17k: Maybe I am missing something, but couldn't you iterate over all recommendations from the page and just check their status to see if they are _selectable_ or not ?

frontend/src/components/CoreTable.vue
Podkamienna: I didn't notice that. It's easy to just show the minus in the right case.

frontend/src/components/CoreTable.vue
Podkamienna: What do you mean by "this"? Is this a suggestion to create some type of which page would have to be? Or do you have something else in mind?

s17k: I meant that I think a name such as `pageChangedHandler` would make what this is doing  more obvious

frontend/src/components/CoreTable.vue
Podkamienna: I really wanted to do it that way, but I couldn't find a way to. If know how to do that, it would be great to change it.

Podkamienna: To be more precise, I couldn't find a way to get  the list of rows that are currently visible.

Podkamienna: Ok, I found it.

frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +121,62 @@ export default class CoreTable extends Vue {
...
+  // Otherwise unselects all rows on the current page
+  selectAll(select: boolean) {
+    select
+      ? this.$store.commit("coreTableStore/selectAllSelectable")
+      : this.$store.commit("coreTableStore/unselectAllSelectable");
+  }
+
+  // Checks if everything is selected on the current page
+  areAllSelected(): boolean {
+    return this.selectableRows.every(this.isSelected);
```
s17k, line 158: I am a bit surprised that this even works: https://stackoverflow.com/questions/30486345/losing-this-context-in-javascript-when-passing-around-members

s17k: If you pass `() => this.isSelected` instead, then `this` will be captured in this anonymous function and it will be a bit clearer why this works.

s17k (changes requested):
otherwise looks good!

s17k (approved):

frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +121,62 @@ export default class CoreTable extends Vue {
...
+  // Otherwise unselects all rows on the current page
+  selectAll(select: boolean) {
+    select
+      ? this.$store.commit("coreTableStore/selectAllSelectable")
+      : this.$store.commit("coreTableStore/unselectAllSelectable");
+  }
+
+  // Checks if everything is selected on the current page
+  areAllSelected(): boolean {
+    return this.selectableRows.every(() => this.isSelected);
```
minorczyka, line 158: I'm not a JS expert but shouldn't it be either `this.selectableRows.every(this.isSelected);` or `this.selectableRows.every(row => this.isSelected(row));`?

frontend/src/components/CoreTable.vue
```diff
@@ -105,12 +121,62 @@ export default class CoreTable extends Vue {
...
+  }
+
+  // Checks if anything is selected on the current page
+  isSomeSelected(): boolean {
+    return this.selectableRows.some(item => this.selectedRows.includes(item));
```
minorczyka, line 163: If you declared `isSelected` function, you should probably use it here.

frontend/src/components/CoreTable.vue
```diff
@@ -18,12 +18,28 @@ limitations under the License. -->
     :headers="headers"
     :items="this.$store.getters['filteredRecommendationsWithExtras']"
     :single-select="false"
-    v-model="selectedRows"
     show-select
+    v-model="selectedRows"
+    v-on:current-items="setSelectableRows"
```
minorczyka, line 23: Using this `setSelectableRows` we're getting a little bit weird dependency. Table depends on recommendatoins from store and store depends on what is displayed in table. But I guess there's no other easy solution for this problem.

minorczyka (changes requested):

frontend/src/components/CoreTable.vue
s17k: Nicely spotted, no wonder I missed it in the review as I actually suggested it...

Podkamienna: I have no idea, why I didn't notice it, as it simply doesn't work...

minorczyka (approved):

PR #122, opened 2020-08-25
add real service

ViktoriiaKovalova:
authorization & listing. /apply was not added yet.


ViktoriiaKovalova:
it's not building because automation package changes are not in master yet. please review `real-service-automation-part` first

cmd/server/apply.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		sendError(c, fmt.Errorf("Authorization code not specified"), http.StatusUnauthorized)
+	}
+}
+
+func getCheckStatusHandler(authService AuthorizationService) func(c *gin.Context) {
+	return func(c *gin.Context) {
+		authCode := c.Request.Header["Authorization"]
+		name := c.Query("name")
+		if len(authCode) != 0 {
```
s17k, line 125: it's not that easy to read 4 nested ifs - the else's are usually far away from conditions.
If you changed this and the next `if` to their opposites (e.g. `len(authCode) == 0`), that would remove the two extra levels.

ViktoriiaKovalova: thanks, fixed

cmd/server/list.go
```diff
@@ -0,0 +1,116 @@
...
+func getListHandler(authService AuthorizationService) func(c *gin.Context) {
+	return func(c *gin.Context) {
+		user, err := authorizeRequest(authService, c.Request)
+
+		if err != nil {
+			sendError(c, err, http.StatusUnauthorized)
+			return
+		}
+
+		handler, loaded := listRequestsInProcess.LoadOrStore(user.email, &listRequestHandler{service: user.service, numConcurrentCalls: defaultNumConcurrentCalls})
```
minorczyka, line 96: Returned `listRequestHandler` is not thread-safe but you're accessing it from multiple threads.

One possible scenario when it won't work is when `h.task` is already incremented to max in `ListAllProjectsRecommendations` but function still haven't returned its result. In L105 you will get empty results from `handler.GetResult()`.

cmd/server/main.go
```diff
@@ -0,0 +1,89 @@
...
+
+	router.GET("/recommendations/checkStatus", getCheckStatusHandler(auth))
+	return router
+}
+
+var listRequestsInProcess listRequestsMap
+var applyRequestsInProcess applyRequestsMap
+
+func main() {
+	if err := setConfig(&config, "config.json"); err != nil {
```
minorczyka, line 73: Reading credentials from file might be troublesome when deploying to AppEngine. We might want to read it from environmental variables or maybe use some dedicated solution for secrets from AppEngine. But let's leave it as it is for now.

cmd/server/apply.go
```diff
@@ -0,0 +1,147 @@
+/*
```
minorczyka, line 1: I haven't reviewed this file because there's the same problem with synchronization as in `list.go` so some changes will be needed.

cmd/server/auth.go
```diff
@@ -0,0 +1,150 @@
...
+		return "", fmt.Errorf("Extracting email failed:" + err.Error())
+	}
+	return claims.Email, nil
+}
+
+func getAuthHandler(authService AuthorizationService) func(c *gin.Context) {
+	return func(c *gin.Context) {
+		authCode := c.Request.Header["Authorization"]
+		if len(authCode) == 0 {
+			sendError(c, fmt.Errorf("Auth code not specified"), http.StatusUnauthorized)
```
minorczyka, line 117: It should be rather BadRequest error here, user was authorized to access this page.

cmd/server/auth.go
```diff
@@ -0,0 +1,150 @@
...
+			return
+		}
+		c.JSON(http.StatusOK, gin.H{"token": token})
+	}
+}
+
+// authorizeRequest extracts the token from Authorization header in request
+// and uses it to return authorized user using authService.
+func authorizeRequest(authService AuthorizationService, request *http.Request) (User, error) {
+	token := request.Header["Authorization"]
```
minorczyka, line 132: Convention for using Authorization header is: `Authorization: <type> <credentials>` (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization).

We don't really have to but we probably should follow this convention and include `Bearer` type.

cmd/server/auth.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(token) == 0 {
+		return User{}, fmt.Errorf("Token not specified")
+	}
+	return authService.Authorize(token[0])
+}
+
+func redirectHandler(c *gin.Context) {
+	email := c.Query("login_hint")
```
minorczyka, line 140: It would be nice to add comment about this query parameter either here or when you're configuring gin's router.

minorczyka (changes requested):

cmd/server/list.go
ViktoriiaKovalova: It is actually done only when result is already assigned(53-54 lines in this file):
```
h.result, h.err = automation.ListAllProjectsRecommendations(h.service, h.numConcurrentCalls, h.task.GetNextSubtask())
h.task.SetAllDone()
```
And GetProgress() will always return (done, all) such that `done < all` until SetAllDone() method was called (from Task implementation).

cmd/server/auth.go
ViktoriiaKovalova: okay

cmd/server/auth.go
ViktoriiaKovalova: yes, I also thought that we should add some type, but i'm not sure how to make it in /auth request, when auth code is sent. Should i also set the type to Bearer or maybe we can pass it as query parameter?

cmd/server/list.go
ViktoriiaKovalova: Do you think I should add some checks in `GetResult()` method also? Right now the best i can think of is adding a comment that it should not be called before `GetProgress()` returns equal values(i.e. `done==all`).

minorczyka: You're right, my example won't happen but modifying not thread-safe object from multiple threads is something you don't want to do no matter what. Even if right now it would work, it will be very easy to introduce some related bugs in the future and race condition bugs are very hard to debug.

minorczyka: I wouldn't expose these objects outside this map, I would build more complex structure for it that would do all necessary synchronization.

cmd/server/auth.go
minorczyka: Oh I overlooked it. But I think in /auth request you won't get authorization code from header but from query parameter anyway.

cmd/server/auth.go
```diff
@@ -0,0 +1,150 @@
...
+		if len(authCode) == 0 {
+			sendError(c, fmt.Errorf("Auth code not specified"), http.StatusUnauthorized)
+			return
+		}
+		token, err := authService.CreateUser(authCode[0])
+		if err != nil {
+			sendError(c, err, http.StatusUnauthorized)
+			return
+		}
+		c.JSON(http.StatusOK, gin.H{"token": token})
```
minorczyka, line 125: It shouldn't return token in JSON, it should redirect user to our frontend with token set in query parameter. But we still don't have handlers for frontend so I guess it can stay like that, just add some comment here.

minorczyka (changes requested):

cmd/server/auth.go
ViktoriiaKovalova: why shouldn't it? it's auth code exchange to token, it doesn't redirect to google login page

minorczyka: Google login page will redirect user's browser to this handler (with authorization code as parameter). If you return token as JSON here, it will just be displayed in user's browser. Instead we should redirect user to page with our frontend and pass token as query parameter.

ViktoriiaKovalova: i don't know if the frontend will handle redirection so i created this /redirect handler which should return the auth code to frontend. The call to /auth for token is not a redirect right now so I am sending token in the body as a response. We can change it to something else, just tell me how it should be done, and i'll update design doc & code.

minorczyka: It won't work but let's merge it as it's now. We will change it later when we will do integration with frontend.

minorczyka (changes requested):

cmd/server/auth.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+// authorizeRequest extracts the token from Authorization header in request
+// and uses it to return authorized user using authService.
+func authorizeRequest(authService AuthorizationService, request *http.Request) (User, error) {
+	token := request.Header["Authorization"]
+	if len(token) != 2 || token[0] != "Bearer" {
```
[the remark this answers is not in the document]
ViktoriiaKovalova: you're right, thanks

cmd/server/auth.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		c.JSON(http.StatusOK, gin.H{"token": token})
+	}
+}
+
+// authorizeRequest extracts the token from Authorization header in request
+// and uses it to return authorized user using authService.
+func authorizeRequest(authService AuthorizationService, request *http.Request) (User, error) {
+	bearToken := request.Header["Authorization"]
+	log.Println(len(bearToken))
```
minorczyka, line 132: Please remove.

minorczyka (approved):

PR #126, opened 2020-08-26
Wait until completion - StartInstance

Podkamienna:
Added waiting until completion of start machine request

### DIFF
--- a/pkg/automation/instances.go
+++ b/pkg/automation/instances.go
@@ -55,6 +55,9 @@ func (s *googleService) StopInstance(project string, zone string, instance strin
 // StartInstance starts instance using instances.start method
 func (s *googleService) StartInstance(project string, zone string, instance string) error {
 	instancesService := compute.NewInstancesService(s.computeService)
-	_, err := instancesService.Start(project, zone, instance).Do()
+	requestID := uuid.New().String()
+	err := AwaitCompletion(func() (*compute.Operation, error) {
+		return instancesService.Start(project, zone, instance).RequestId(requestID).Do()
+	})
 	return err
 }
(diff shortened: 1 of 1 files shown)


pkg/automation/instances.go
```diff
@@ -55,6 +55,9 @@ func (s *googleService) StopInstance(project string, zone string, instance strin
 // StartInstance starts instance using instances.start method
```
s17k, line 55: It is obvious that StartInstance starts some instance, I would focus on stuff that is hardest to guess from the code.
Like, what permissions are required (very often mentioned in other googleService methods here).

Podkamienna: The comment is exactly the same as the one in the StopInstance

s17k: So probably good to change that one too!
https://github.com/googleinterns/recomator/blob/master/pkg/automation/disks.go has nice comments, for example

Podkamienna: As I was doing some part of the project in the project written mainly by Viktoriia, I decided to follow the convention she chose. The permissions are listed in access.go - and I think that the part of making the code modular is not having to think about them in different places. So I think, that mentioning them would only confuse the person wanting to just check, what is the function for. Moreover, from my experience this kind of comment may also be helpful - especially to someone not familiar with the API.

s17k: If you want to follow Viktoriia's conventions, the permissions are actually written down above function definitions in many files. But I wouldn't say that following conventions is the end goal.

Podkamienna: Okay, then I can change all the comments in instances.go to also mention permissions.

s17k:
LGTM, but I don't know the context for this change so probably best if Viktoriia/Artur review it first.

ViktoriiaKovalova (approved):
LGTM

PR #129, opened 2020-08-27
Frontend - more flexible fetch & better handling of status updates

s17k:
- Existing recommendations are removed at the beginning of a fetch now (so that if we have authorization errors, we can easily and safely restart it):  https://github.com/googleinterns/recomator/pull/129/commits/106045e7fa1e10fe237403633e72a71808ae6c44
- We now track recommendations which need to have their status watched with a new field in `RecommendationExtra` called `needsStatusWatcher`. This allows us to just iterate through all recommendations in the watchStatus infinite loop, and easily handle the recommendations list changing. 

Requests (/apply, /watchStatus) are now sent sequentially (waiting asynchronously for the previous one to finish before sending a new one). What used to be done was sending all the requests one after another, then asynchronously handling all their responses. This used to cause the UI to halt for a few seconds and update much slower (most probably because there was a large chunk of work that stayed at the top of the event queue for these few seconds).
 https://github.com/googleinterns/recomator/pull/129/commits/b085b2a88fc67688b1048b8d5fc62d9622ba0234 
https://github.com/googleinterns/recomator/pull/129/commits/d9263c385ffd3bdf97b111ef26b07e7ed467fbc4 (second commit ammends the first one so that requests are done sequentially, as described above) 
https://github.com/googleinterns/recomator/pull/129/commits/54ff19f2150b440ea2d50407b1830d051af87c91 (made a function out of the core of `centralStatusWatcher`) 
https://github.com/googleinterns/recomator/pull/129/commits/3e2821e6173b46e2acf7f94e03a44f3c093b48bc (tests)

  - recommendations that were `CLAIMED` when we got them are no longer watched (at least for now, added this as an issue: https://github.com/googleinterns/recomator/issues/131), as this makes it complex to distinguish between getting `NOT APPLIED` from `checkStatus` because:
     - the backend hasn't acknowledged the `apply` request that the frontend has sent
     - the recommendation was applied earlier and is still `CLAIMED`, for example by Pantheon or an API or even the same Recomator backend instance but in a different fetch
```typescript
async startCentralStatusWatcher({ commit, dispatch }): Promise<void> {
    commit("registerCentralStatusWatcher");
    for (;;) {
      await dispatch("checkStatusOnceForAll");
      // ask the browser to do something else for a bit and then resume
      await delay(APPLY_PROGRESS_WAIT_TIME);
    }
 }
async checkStatusOnceForAll({ state, commit, dispatch }): Promise<void> {
    // make a copy first to make the array constant inside the for loop
    const recsCopy = state.recommendations.map(rec => rec);

    // check status of all recommendations that need to be watched
    // (waits for a response before starting a new request)
    for (const rec of recsCopy) {
      if (!rec.needsStatusWatcher) continue;

      const shouldContinue = await dispatch("checkStatusOnce", rec);
      commit("setRecommendationNeedsStatusWatcher", {
        recName: rec.name,
        needs: shouldContinue
      });
    }
 }
```
- Fix: status was changed directly, not through a mutation https://github.com/googleinterns/recomator/pull/129/commits/2b56fa2ea37b0352f25ddb16f7b3b41617a7fb73
- Fix: status filter options were changing very frequently (more than once per second) when statuses were updated https://github.com/googleinterns/recomator/pull/129/commits/5bfa97aca4de90578f311fa2b01586a895b68282
- Performance fix: changing `v-if` to `v-show` improved rendering performance significantly (from what I observed visually), as I observed that icons render for the longest time (Vue extension shows these stats) so it is better not to re-render them each time status changes (`v-if`'s behaviour). https://github.com/googleinterns/recomator/pull/129/commits/02d79f923a9407c15a05e2da8d9e1cfc0278ff47

### DIFF
--- a/frontend/src/store/recommendations.ts
+++ b/frontend/src/store/recommendations.ts
@@ -215,30 +215,33 @@ const actions: ActionTree<IRecommendationsStoreState, IRootStoreState> = {
   //
   // It is safe to add to/clear the recommendations list during execution,
   // as we are operating on a copy
-  async startCentralStatusWatcher({ state, commit, dispatch }): Promise<void> {
+  async startCentralStatusWatcher({ commit, dispatch }): Promise<void> {
     commit("registerCentralStatusWatcher");
     for (;;) {
-      // make a copy first so that this is more predictable
-      const recsCopy = state.recommendations.map(rec => rec);
-
-      // check status of all recommendations that need to be watched
-      // (waits for a response before starting a new request)
-      for (const rec of recsCopy) {
-        if (!rec.needsStatusWatcher) continue;
-
-        const shouldContinue = await dispatch("watchStatusOnce", rec);
-        commit("setRecommendationNeedsStatusWatcher", {
-          recName: rec.name,
-          needs: shouldContinue
-        });
-      }
+      await dispatch("checkStatusOnceForAll");
       // ask the browser to do something else for a bit and then resume
       await delay(APPLY_PROGRESS_WAIT_TIME);
     }
   },
+  async checkStatusOnceForAll({ state, commit, dispatch }): Promise<void> {
+    // make a copy first to make the array constant inside the for loop
+    const recsCopy = state.recommendations.map(rec => rec);
+
+    // check status of all recommendations that need to be watched
+    // (waits for a response before starting a new request)
+    for (const rec of recsCopy) {
+      if (!rec.needsStatusWatcher) continue;
+
+      const shouldContinue = await dispatch("checkStatusOnce", rec);
+      commit("setRecommendationNeedsStatusWatcher", {
+        recName: rec.name,
+        needs: shouldContinue
+      });
+    }
+  },
   // Should return nearly immediately. Assumes the recommendation was applied by us (not Pantheon, for example)
   // the promise encapsulates the answer to: do we want to continue watching this recommendation?
-  async watchStatusOnce(
+  async checkStatusOnce(
     { commit },
     rec: RecommendationExtra
   ): Promise<boolean> {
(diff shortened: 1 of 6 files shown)


frontend/src/store/recommendations.ts
```diff
@@ -192,26 +204,39 @@ const actions: ActionTree<IRecommendationsStoreState, IRootStoreState> = {
     }
   },
   // If there is a "CLAIMED" recommendation, its status will change once it
-  //  has finished being applied. Therefore, we want to follow it and update the store
-  //  (which will, in turn, automatically update the UI).
-  async watchStatus({ dispatch }, rec: RecommendationExtra): Promise<void> {
+  // has finished being applied. Therefore, we want to follow it and update the store
+  // (which will, in turn, automatically update the UI).
+  //
+  // It is safe to add to/clear the recommendations list during execution,
```
s17k, line 210: Maybe it was not such a good idea to put `await`s inside a forEach after all.
From what I understand, it doesn't actually handle the requests one by one. It only sends them one by one.

minorczyka (changes requested):
Sorry that I put comments instead of review.

s17k:
> Are you sure it's fine? `forEach` uses async function so I can imagine that `state.recommendations` can be modified when `forEach` waits for `await dispatch("watchStatusOnce", rec)`.

This has been resolved already, as we discussed.

> It's not obvious that this is never ending loop. If it's meant to be called only once maybe it shouldn't be store's action and just a normal function.

https://github.com/googleinterns/recomator/pull/129/commits/ed0f471bd52db2c491ed181656e0974b02596b49

s17k:
Just letting you know that I have added the tests and this PR should be complete and ready to review now.

minorczyka (approved):

Issue #124, opened 2020-08-25
Handle the 'I can't find the recommendation I selected' problem somehow.


s17k:
There were two alternatives proposed up to now:
- 'X' button in the footer, which removes all selections
- filtering 'Selected'/'Unselected' in the filters row

PR #134, opened 2020-08-28
Footer upgrade

s17k:
- adds the 'X' button that @minorczyka suggested to clear selections and hide the footer (closes: #124)
- in my opinon a blue footer looks nicer than a yellow one; that is the primary theme colour and the colour of the header. I am open to suggestions though.
- the UX expert suggested to keep lines below 70 characters, so the summary occupies 2 or 3 lines now (closes: #103)
- simplified the margins/paddings: having too many screen size breakpoints will be hard to maintain 

### Screenshots (before/after, desktop/mobile)
**Desktop before**:

**Desktop after**:

**Mobile before**:

**Mobile after**:

### DIFF
--- a/frontend/src/components/Footer.vue
+++ b/frontend/src/components/Footer.vue
@@ -17,24 +17,26 @@ limitations under the License. -->
     <v-footer
       v-if="selectedRowsCount > 0"
       v-bind:fixed="true"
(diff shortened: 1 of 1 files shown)


brzezinski (approved):
LGTM

PR #142, opened 2020-09-01
Status filter options made constant: fixed #140

### LINKED ISSUE #140: Fix status filter options (make them constant)
What I observed was that page size was the factor determining the speed of statuses updates in the UI.

They are currently recalculated after each status change. That is why, if you open about a 1000 of recommendations and apply all of them, they will need to be recalculated about 1000 times. This combined with frequent repainting makes it unacceptably slow.

s17k:
What I observed was that page size was the factor determining the speed of statuses updates in the UI.

Status filter options are currently dynamically calculated from all observed statuses (possibly thousands of them). That is why, if you open about a 1000 of recommendations and apply all of them, these options will need to be recalculated about 1000 times (possibly millions of operations). This combined with frequent repainting makes it unacceptably slow.

I fixed them to all internal statuses instead, the statuses are now updating at least 2x faster (possibly much more than that) for larger page sizes.

### DIFF
--- a/frontend/src/components/filters/StatusFilter.vue
+++ b/frontend/src/components/filters/StatusFilter.vue
@@ -14,7 +14,7 @@ limitations under the License. -->
 <template>
   <v-combobox
     v-model="statusesSelected"
-    :items="allStatuses"
+    :items="allInternalStatuses"
     label="Select status"
     multiple
   >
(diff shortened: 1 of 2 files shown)


frontend/src/components/filters/StatusFilter.vue
```diff
@@ -36,8 +37,8 @@ export default class StatusFilter extends Vue {
     this.$store.commit("coreTableStore/setStatusesSelected", statuses);
   }
 
-  get allStatuses(): string[] {
-    return this.$store.getters["recommendationsStore/allStatuses"];
+  get allInternalStatuses(): string[] {
+    return allInternalStatuses();
```
brzezinski, line 41: please put any comment what's going on here. User should know what we're doing so there is no need to see allInternalStatuses() definition.

brzezinski (approved):
LGTM

PR #144, opened 2020-09-02
add time parameter

ViktoriiaKovalova:


minorczyka (approved):

pkg/automation/disks.go
```diff
@@ -51,6 +52,11 @@ func randomSnapshotName(zone string, disk string, generator *rand.Rand) (string,
 	return result, nil
 }
 
+const (
```
s17k, line 55: I would personally put all of these together in `service.go`, closer to the place where they are actually used and so that it is easier to figure out what is the sleepTime actually for and how do they compare with each other. But it is really not that important.

s17k (approved):

Issue #153, opened 2020-09-03
Add a test for Pantheon links when type==INCREASE_PERFORMANCE, once we integrate with the real service.


s17k:
The fake service doesn't have INCREASE_PERFORMANCE recommendations for now, so there is no easy way to get a sample recommendation of this type in JSON form.

PR #154, opened 2020-09-03
Frontend resource link

s17k:
Adds Pantheon links for instances and disks. Closes: #152 
It is not that easy to make a test for Pantheon links for "INCREASE_PERFORMANCE" recommendations yet (not sent by the fake service), so I have created an issue to remember about this once we integrate with the real service when this will be much easier (#153).

First four commits are small tweaks to the related code, the last one contains the actual feature.

### DIFF
--- a/frontend/src/components/ResourceCell.vue
+++ b/frontend/src/components/ResourceCell.vue
@@ -13,13 +13,14 @@ See the License for the specific language governing permissions and
 limitations under the License. -->
 <template>
   <div>
-    {{ shortName }}
+    <a :href="pantheonLink"> {{ shortName }} </a>
   </div>
 </template>
 <script lang="ts">
 import Vue, { PropType } from "vue";
 import { Component } from "vue-property-decorator";
 import { RecommendationExtra } from "../store/data_model/recommendation_extra";
+import { getResourcePantheonLink } from "../store/data_model/recommendation_raw";
 
 const ResourceCellProps = Vue.extend({
   props: {
@@ -34,6 +35,9 @@ const ResourceCellProps = Vue.extend({
 export default class ResourceCell extends ResourceCellProps {
   get shortName() {
     return this.rowRecommendation.resourceCol;
+  }
+  get pantheonLink() {
+    return getResourcePantheonLink(this.rowRecommendation);
   }
 }
 </script>
--- a/frontend/src/components/ResourceCell.vue
+++ b/frontend/src/components/ResourceCell.vue
@@ -13,14 +13,14 @@ See the License for the specific language governing permissions and
 limitations under the License. -->
 <template>
   <div>
-    <a :href="pantheonLink"> {{ shortName }} </a>
+    <a :href="consoleLink"> {{ shortName }} </a>
   </div>
 </template>
 <script lang="ts">
 import Vue, { PropType } from "vue";
 import { Component } from "vue-property-decorator";
 import { RecommendationExtra } from "../store/data_model/recommendation_extra";
-import { getResourcePantheonLink } from "../store/data_model/recommendation_raw";
+import { getResourceConsoleLink } from "../store/data_model/recommendation_raw";
 
 const ResourceCellProps = Vue.extend({
   props: {
@@ -36,8 +36,8 @@ export default class ResourceCell extends ResourceCellProps {
   get shortName() {
     return this.rowRecommendation.resourceCol;
   }
-  get pantheonLink() {
-    return getResourcePantheonLink(this.rowRecommendation);
+  get consoleLink() {
+    return getResourceConsoleLink(this.rowRecommendation);
   }
 }
 </script>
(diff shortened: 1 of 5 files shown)


frontend/src/components/ResourceCell.vue
```diff
@@ -13,13 +13,14 @@ See the License for the specific language governing permissions and
 limitations under the License. -->
 <template>
   <div>
-    {{ shortName }}
+    <a :href="pantheonLink"> {{ shortName }} </a>
```
brzezinski, line 16: please do not refer to internal console name Pantheon. Instead please use "console"

frontend/src/components/ResourceCell.vue
```diff
@@ -13,13 +13,14 @@ See the License for the specific language governing permissions and
...
   </div>
 </template>
 <script lang="ts">
 import Vue, { PropType } from "vue";
 import { Component } from "vue-property-decorator";
 import { RecommendationExtra } from "../store/data_model/recommendation_extra";
+import { getResourcePantheonLink } from "../store/data_model/recommendation_raw";
```
brzezinski, line 23: same as above. Avoid "pantheon"

frontend/src/store/data_model/recommendation_raw.ts
```diff
@@ -137,6 +136,12 @@ export function getRecommendationFirstResource(
   return recommendation.content.operationGroups[0].operations[0].resource;
 }
 
+export function getRecommendationSecondResource(
+  recommendation: RecommendationRaw
+): string {
+  return recommendation.content.operationGroups[0].operations[1].resource;
```
brzezinski, line 142: is there a chance there will be no second operation? any error handling?

frontend/src/store/data_model/recommendation_raw.ts
```diff
@@ -150,6 +155,24 @@ export function getRecommendationFirstValue(
   return recommendation.content.operationGroups[0].operations[0].value;
 }
 
+// (b, /a/b/c/d/e) => c
+export function extractFromResource(
```
brzezinski, line 159: any comment would be nice to have what are you extracting here

frontend/src/store/data_model/recommendation_raw.ts
```diff
@@ -232,3 +255,29 @@ export function getRecommendationCostPerWeek(
...
+}
+
+export function getResourcePantheonLink(recommendation: RecommendationRaw) {
+  const zone = getRecommendationZone(recommendation);
+  const project = getRecommendationProject(recommendation);
+  const shortName = getRecommendationResourceShortName(recommendation);
+  const type = getRecommendationType(recommendation);
+  switch (type) {
+    case "SNAPSHOT_AND_DELETE_DISK": // disks
+      return `https://pantheon.corp.google.com/compute/disksDetail/zones/${zone}/disks/${shortName}?project=${project}`;
```
brzezinski, line 278: it's internal link. outside world link is console.cloud.google.com

frontend/tests/unit/recommendations.spec.ts
```diff
@@ -40,17 +46,45 @@ describe("Store", () => {
...
+test("Getting the zone of the resource", () => {
+  expect(getRecommendationZone(freshSampleRawRecommendation())).toEqual(
+    "us-east1-b"
+  );
+});
+
+describe("Getting a Pantheon link for the resource", () => {
+  test("type: CHANGE_MACHINE_TYPE ", () => {
+    expect(getResourcePantheonLink(freshSampleRawRecommendation())).toEqual(
+      "https://pantheon.corp.google.com/compute/instancesDetail/zones/us-east1-b/instances/alicja-test?project=rightsizer-test"
```
brzezinski, line 70: this link will only work in corp network

brzezinski (changes requested):
please fix

frontend/src/store/data_model/recommendation_raw.ts
s17k: There is one comment already in line 158, but I will add something more meaningful.

frontend/src/store/data_model/recommendation_raw.ts
s17k: It implicitly assumes that if you are asking about the second operation, you assume that it exists.
I chose not to make an error handler here because:
- we need to assume some structure of the data that we are given (for example, that no one will ask about the second resource if they did not previously check for recomender type). Returning null sometimes in some scenarios would cause more problems than it would solve; all data would be accepted and we would never probably find out about some slight inconsistencies (as the app would be always quietly handling all broken data).
- for consistency: the `getRecommendationFirstResource function` (written earlier) did not have one

s17k:
Oh wow, gj catching all of these. I had no idea Pantheon was not an official name.

brzezinski (approved):
LGTM

PR #155, opened 2020-09-07
Frontend tests

Podkamienna:
I couldn't test the checkbox of the v-data-table, because its behavior when tested was unexpected.


s17k:
Could you please elaborate on what you said about the checkbox behaviour being unexpected ? Was it just not responding to data changes ?

Podkamienna:
The problem was that the checkbox didn't react to the events generated by jest (for example clicking on the field didn't make it checked, when it should have).

s17k:
Right, this might be due to a few issues that I know of:
- Checkboxes don't work if they don't have a v-model of their own or something equivalent.
- You might need to tell Vue explicitly to update the DOM using [Vue.nextTick() or something else](https://vue-test-utils.vuejs.org/guides/testing-async-components.html)
- Once you emit a click event, the function that actually does that might stay at the top of the microtasks queue, so some stuff that would normally run won't (mentioned somewhere near the end of [this](https://www.youtube.com/watch?v=cCOL7MC4Pl0)

s17k (commented):
~~We should not merge files by committing the changes again, as in the second commit: 9ce7e7a061b044fb3a571ec5e84af605501c828f
It is much cleaner to use `git merge master` so that it is obvious where the code was introduced. 
Otherwise, the commit history for a file will be really hard to read.~~
(Although, I now see that these were changes from the future that you needed. Probably easier would be to create a branch from that branch that wasn't merged yet and build upon these unmerged commits. Github is smart enough to hide them in the new PR once they are merged).
~~Also, this way you will not have it displayed as a change on Github so it will be much easier to focus on what you are actually modifying (as it is not really a new change, just a merge of previous changes).~~

frontend/tests/unit/sample_recommendations.ts
```diff
@@ -0,0 +1,165 @@
...
+};
+
+export function freshSampleRawRecommendation(): RecommendationRaw {
+  // deep copy
+  return JSON.parse(
+    JSON.stringify(sampleRawRecommendation)
+  ) as RecommendationRaw;
+}
+
+export function freshSavingRawRecommendation(): RecommendationRaw {
```
s17k, line 153: In the latest version of this file, a few of these are already introduced: https://github.com/googleinterns/recomator/blob/master/frontend/tests/unit/sample_recommendation.ts
Best if you re-use the deep copy function implemented there.

s17k:
If you could please not change the name of the file: sample_recommendation.ts to sample_recommendations.ts and modify it in the same PR. If you just modify it, it will be much much easier to compare the changes on Github (otherwise Github treats this as deleting one file and creating another)

PR #159, opened 2020-09-08
Similarity-based model - simplified version

s17k:
- As described in the design doc, but without taking costs/savings into account (we can maybe introduce it later at some point)
- Recommendations are marked as applied in training data at the moment someone clicks 'Apply' or selects them and clicks 'ApplySelected'
- I considered adding them only when they succeed (so that retried recommendations are not counted multiple times, but this causes a new problem: we would need to wait for the status to update before adding a recommendation to training data and someone might just close the tab before that happens). This should not be a problem anyway, as we do not anticipate a large number of retried recommendations.


Podkamienna:
I actually think that it is better to have both succeeded and failed recommendations in the training data.

frontend/src/store/smart_sort/similarity.ts
```diff
@@ -0,0 +1,105 @@
...
+    data.typeCounters,
+    rec.typeCol
+  )[0];
+  const partForType = 1 - appliedWithSameType / appliedSize;
+
+  return partForProject + partForType;
+}
+
+// Orders (in place) by average similarity to applied recommendations.
+// See the design doc for details
```
brzezinski, line 52: users won't have access to desing doc details

brzezinski (approved):
one small comment

PR #161, opened 2020-09-09
Changed the address

Podkamienna:
Added a temporary address for production, which is needed for github pages to work.


frontend/src/config.ts
```diff
@@ -14,5 +14,6 @@ limitations under the License. */
 
 export class ProjectConfig {
   public static DEVELOPMENT_SERVER_ADDRESS = "http://localhost:8000";
-  public static PRODUCTION_SERVER_ADDRESS = "";
+  public static PRODUCTION_SERVER_ADDRESS =
+    "https://recomator-282910.ey.r.appspot.com";
```
minorczyka, line 18: Instead of hard-coding it here, I would prefer to just use relative paths in production mode.

Podkamienna: Could you explain, what do you mean by using relative paths? Because I am not sure I understand. But an idea that I have is having another config.ts on the branch pages and not changing this file in master (excluding it in the workflow file).

minorczyka (approved):

PR #164, opened 2020-09-09
Frontend auth

s17k:
Closes: #164 (Frontend auth)
This also fixes: #108 as the white space at the top was caused by the router link being hidden behind the top bar.


s17k:
Some commits are duplicated somewhere else. I don't like force-pushing so I am closing this.

PR #168, opened 2020-09-11
Frontend move fetch

s17k:
- Starting fetching and status watchers has now been moved to router/index.ts so that they only launch when we are in the / route. This was not a problem when we had only one component, but we are about to add the permissions page which will need a new one.
- Removed the router link to / from each page (which caused the white space at the top, Fixes: #108)

### DIFF
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -15,6 +15,7 @@ limitations under the License. */
 import Vue from "vue";
 import VueRouter, { RouteConfig } from "vue-router";
 import Home from "../views/Home.vue";
+import store from "../store/root";
 
 Vue.use(VueRouter);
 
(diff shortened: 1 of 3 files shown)


frontend/src/router/index.ts
```diff
@@ -15,14 +15,22 @@ limitations under the License. */
 import Vue from "vue";
 import VueRouter, { RouteConfig } from "vue-router";
 import Home from "../views/Home.vue";
+import store from "../store/root";
```
minorczyka, line 18: nit: use `"@/store/root"`

frontend/src/router/index.ts
```diff
@@ -15,14 +15,22 @@ limitations under the License. */
...
   {
     path: "/",
     name: "Home",
-    component: Home
+    component: Home,
+    beforeEnter(_, __, next) {
+      // Asynchronously request and receive recommendations from the middleware
+      store.dispatch("recommendationsStore/fetchRecommendations");
+      // Start status watchers
+      store.dispatch("recommendationsStore/startCentralStatusWatcher");
```
minorczyka, line 31: We will probably have to move them again later. When we add another page (with project selection and permission tests or with graphs) we for sure won't want to start status watchers again and list recommendations maybe if we changed projects. But it can stay for a while.

minorczyka (approved):

frontend/src/router/index.ts
s17k: Yes. We can add an if, to only do that if we just redirected from the auth code receiver endpoint.

PR #169, opened 2020-09-11
Frontend fix cost parsers

s17k:
- "improve performance" recommendations have their cost computed properly now
- there used to be no tests for cost parsers, so added two for recommendations with cost in primaryImpact and additionalImpact respectively


brzezinski (approved):
LGTM

PR #175, opened 2020-09-11
Frontend: auth

s17k:
~~Auth integration is waiting for requestId to be implemented in the backend.
In the meantime, we can review the auth code that has been written so far!~~
Authorization added to frontend, uses requestIds (as defined in the design doc) to track selected projects.

### DIFF
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -16,6 +16,8 @@ import Vue from "vue";
 import VueRouter, { RouteConfig } from "vue-router";
 import Home from "../views/Home.vue";
 import store from "../store/root";
+import { IRootStoreState } from "../store/root";
+import { getBackendAddress } from "../config";
 
 Vue.use(VueRouter);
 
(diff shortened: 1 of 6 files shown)


frontend/src/store/auth.ts
```diff
@@ -0,0 +1,69 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */
+
+import { Module, MutationTree } from "vuex";
+import { IRootStoreState } from "./root";
+import store from "@/store/root";
```
minorczyka, line 17: I've only now realized that we have circular dependencies in stores (core_table.ts and recommendations.ts also have them). I'm surprised that it works, I remember times when it caused a lot of troubles with JS. 

But even if it works, circular dependencies should be avoided. It would be great if we removed it but let's not do it in this PR.

frontend/src/router/index.ts
```diff
@@ -25,11 +27,72 @@ const routes: Array<RouteConfig> = [
...
+    // receive the authCode, exchange it for a token and finally
+    // fetch recommendations and go to "/" with a saved token
+    path: "/auth",
+    name: "GetTokenAndGoHome",
+    // nothing will be rendered until next() is called, so it is fine
+    // that this function is asynchronous
+    async beforeEnter(to, _, next) {
+      const authCode = to.query.code;
+      if (authCode == undefined) {
+        next(new Error("The auth code is missing"));
```
minorczyka, line 58: What happens when you pass error to next?

frontend/src/router/index.ts
```diff
@@ -25,11 +27,72 @@ const routes: Array<RouteConfig> = [
...
+
+      store.commit("authStore/setIDToken", token);
+      next({ name: "HomeWithInit" });
+    }
+  },
+  {
+    // Show the main page with recommendations, but initialize first
+    path: "/homeWithInit",
+    name: "HomeWithInit",
+    beforeEnter(_, __, next) {
```
minorczyka, line 88: I wonder what will happen when user manually navigates to this route when we're already fetching recommendations.

s17k: Probably some error. It shouldn't matter, this is not an actual user route that someone is supposed to visit. I only put this as a route and not a function, because it makes the code more local and easier to follow I think.
I could make the path "homeWithInitPleaseDontVisit", it wouldn't change a thing.

frontend/src/router/index.ts
s17k: It will abort the navigation and throw this error (I tested it and saw it in the browser console).
I considered just throwing, but I don't want the router to be smart and redirect back to the `from` address, so that is what this is for. 

Router docs say: "if the argument passed to next is an instance of Error, the navigation will be aborted and the error will be passed to callbacks registered via router.onError()."

https://router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards

frontend/src/store/auth.ts
s17k: Yes! I have just made an issue for that: https://github.com/googleinterns/recomator/issues/180

Podkamienna (approved):

frontend/src/router/index.ts
minorczyka: We should register to this router.onError and display some page with error message. Showing errors only in console is perfect for debugging but it's not the best for user experience :smile: 

Please do this in next PR.

minorczyka (approved):

frontend/src/router/index.ts
s17k: Yes, I totally agree. We lack dialogs for errors all over the place. For example, fetch only saves errors to a store variable that we never show. I have just created an issue for this: #188

PR #177, opened 2020-09-14
add checking permissions endpoint and request ids

ViktoriiaKovalova:


cmd/server/requirements.go
```diff
@@ -0,0 +1,136 @@
...
+			return
+		}
+
+		handler := NewCheckRequestHandler(user.service, checkRequest.Projects)
+		lengthOfID := 20
+		requestID := randomString(lengthOfID, rand.New(rand.NewSource(time.Now().UnixNano())))
+
+		for {
+			err := service.requests.StartProcessing(RequestInfo{user.email, requestID}, handler)
+			if err != nil {
```
minorczyka, line 102: A word of comment for this loop would be nice. Let's say that `StartProcessing` can only fail in case of requestID collision.

minorczyka (approved):

PR #191, opened 2020-09-16
Fix: Close after 'retry' button clicked

s17k:
Fixes: #186

### DIFF
--- a/frontend/src/components/ApplyAndStatusCell.vue
+++ b/frontend/src/components/ApplyAndStatusCell.vue
@@ -63,7 +63,14 @@ limitations under the License. -->
         </v-card-text>
         <v-card-actions>
           <v-spacer></v-spacer>
-          <v-btn color="primary" dark v-on:click="applyRecommendation()">
+          <v-btn
+            color="primary"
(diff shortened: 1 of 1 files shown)


frontend/src/components/ApplyAndStatusCell.vue
```diff
@@ -63,7 +63,14 @@ limitations under the License. -->
         </v-card-text>
         <v-card-actions>
           <v-spacer></v-spacer>
-          <v-btn color="primary" dark v-on:click="applyRecommendation()">
+          <v-btn
+            color="primary"
+            dark
+            v-on:click="
+              errorDialogOpened = false;
```
minorczyka, line 70: nit: I think you could put it into the `applyRecommendation` function.

minorczyka (approved):

frontend/src/components/ApplyAndStatusCell.vue
s17k: I wanted applyRecommendation to have a single responsibility, as it is also used as an 'apply' handler.

PR #195, opened 2020-09-17
Frontend project selection

Podkamienna:


frontend/src/components/CoreTable.vue
```diff
@@ -160,7 +160,10 @@ export default class CoreTable extends Vue {
 
   // Checks if everything is selected on the current page
   areAllSelected(): boolean {
-    return this.selectableRows.every(item => this.isSelected(item));
+    return (
+      this.selectableRows.every(item => this.isSelected(item)) &&
+      this.selectableRows.length !== 0
+    );
   }
```
s17k, line 167: I couldn't guess why this has been added. Could you explain ?

frontend/src/components/ProjectList.vue
```diff
@@ -0,0 +1,165 @@
...
+      <v-tooltip top transition="none">
+        <template v-slot:activator="{ on, attrs }">
+          <v-btn
+            v-on="on"
+            v-bind="attrs"
+            color="secondary"
+            style="font-weight: bold"
+            class="ma-2"
+            rounded
+            depressed
```
s17k, line 78: :(

s17k, line 78: :(

frontend/src/components/ProjectsWithRequirements.vue
```diff
@@ -0,0 +1,178 @@
...
+</style>
+
+<script lang="ts">
+import { Component, Vue } from "vue-property-decorator";
+import { IRootStoreState } from "../store/root_state";
+import { ProjectRequirement } from "../store/data_model/project_with_requirements";
+
+@Component({})
+export default class ProjectList extends Vue {
+  requirementList = [
```
s17k, line 114: Is there a difference between these and the headers ? If not, why not only store these once ?

Maybe move these to a ts file somewhere in the store ? Importing from Vue files is not possible/is an unnecessary pain and someone might want to reuse. It would also be nice to just have them in a nice small file.

frontend/src/config.ts
```diff
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 export class ProjectConfig {
-  public static DEVELOPMENT_BACKEND_ADDRESS = "http://localhost:8078";
+  public static DEVELOPMENT_BACKEND_ADDRESS = "http://localhost:8000";
```
s17k, line 16: nit: we don't need to change this

frontend/src/store/requirements_store.ts
```diff
@@ -0,0 +1,171 @@
...
+limitations under the License. */
+
+import { ProjectRequirement } from "@/store/data_model/project_with_requirements";
+import { delay } from "./utils/misc";
+import { Module, MutationTree, ActionTree } from "vuex";
+import { IRootStoreState } from "./root_state";
+import { Requirement } from "./data_model/project_with_requirements";
+import { getBackendAddress } from "@/config";
+import { getAuthFetch } from "./auth/auth_fetch";
+import router from "@/router";
```
s17k, line 22: This makes a circular dependency.
Router already imports the root store which imports this store, so you can use the router reference saved in the root store.

frontend/src/main.ts
```diff
@@ -20,8 +20,6 @@ import vuetify from "./plugins/vuetify";
 
 Vue.config.productionTip = false;
 
-store.state.router = router;
-
 new Vue({
```
s17k, line 23: Please don't delete that, it is there so that we don't import the router in store files (which are... imported by the router).

s17k (changes requested):

frontend/src/main.ts
Podkamienna: Ok, I deleted that, because I didn't understand what it was doing and I couldn't build the project with it.

frontend/src/components/ProjectsWithRequirements.vue
Podkamienna: Yes, I plan to move them. There is a task for it in the backlog actually. But I think we decided that this is not very urgent.

frontend/src/router/index.ts
```diff
@@ -33,6 +35,33 @@ const routes: Array<RouteConfig> = [
...
+    component: Requirements,
+    beforeEnter(_, __, next) {
+      // Asynchronously request and receive requirements from the middleware
+      store.dispatch("requirementsStore/fetchRequirements");
+      next();
+    }
+  },
+
+  {
+    path: "/recommendations",
```
s17k, line 56: It is hard to figure out just by name what is the difference between /recommendations and / routes (once you know that / is the table with recommendations).
I also don't get why is this a new route, please just move these back to HomeWithInit. If you don't want to start the central status watcher twice, just check if it exists first and only start it if it doesn't. That should be fine.

s17k (changes requested):

frontend/src/store/projects_store.ts
```diff
@@ -0,0 +1,101 @@
...
+    context.commit("startFetch");
+
+    const authFetch = getAuthFetch(context.rootState);
+
+    // First, select the projects
+    const response = await authFetch(`${BACKEND_ADDRESS}/projects`);
+    const responseCode = response.status;
+
+    if (responseCode !== HTTP_OK_CODE) {
+      context.commit("setError", {
```
s17k, line 73: setError is not a mutation

frontend/src/router/index.ts
```diff
@@ -33,6 +35,33 @@ const routes: Array<RouteConfig> = [
         next({ name: "GoogleSignIn" });
         return;
       }
+
+      store.dispatch("projectsStore/fetchProjects");
+      next();
+    }
```
s17k, line 41: Can't we just fetch projects when we need them ? If we are already entering this page, we must have already done some project selection anyway.

frontend/src/router/index.ts
Podkamienna: But why is HomeWithInit necessary?

frontend/src/store/recommendations_store.ts
```diff
@@ -163,12 +164,15 @@ const actions: ActionTree<IRecommendationsStoreState, IRootStoreState> = {
       await delay(FETCH_PROGRESS_WAIT_TIME);
     }
 
-    for (const recommendation of responseJson.recommendations)
-      context.commit("addRecommendation", recommendation);
-    for (const recommendation of context.state.recommendations)
-      trainingDataHandler.addRecommendation(recommendation);
+    if (responseJson.recommendations !== null) {
```
s17k, line 167: I remember that the backend returns a null instead of [] so this is actually needed unforuntately. It would be good to add this as an issue in the backend and put a TODO here to remove this once it is resolved.

frontend/src/router/index.ts
Podkamienna: I don't understand why do you think so?

frontend/src/store/recommendations_store.ts
Podkamienna: I am not sure, but maybe an empty array in Go is the same as nil.

frontend/src/router/index.ts
s17k: Let's imagine that we add some graphs or anything that doesn't actually require a reload of recommendations.
Then, we would like to return to Home without initialising recommendations and starting the status watcher. So instead of redirecting to HomeWithInit, they can redirect directly to Home.
Now, in the case of projects/requirements, we always want to reload all the recommendations so it might actually seem like this between Home and HomeWithInit might be unnecessary.

frontend/src/store/recommendations_store.ts
Podkamienna: Ok, empty slice should be encoded as [] in JSON

Podkamienna: I guess an empty slice shold be encoded as [] is JSON.

frontend/src/router/index.ts
s17k: / (Home) is to be entered in two cases:
- someone just typed our address, so if they don't have a token, let's redirect to Google Sign in
- we have already initialised everything (for example, in HomeWithInit) and we want to show it (either for the first time, or maybe again if we have just shown some graphs)

frontend/src/router/index.ts
Podkamienna: But in that case we would probably have a Home page, where you can select if you want to see some graphs or recommendations.

frontend/src/store/recommendations_store.ts
s17k: Fair point, that is indeed how seems to be.
Imo the backend should fix it (as if we fix it in the backend not frontend, we will have a simpler interface) but it's not a huge deal. I would still put it in the backlog.

frontend/src/router/index.ts
Podkamienna: So in my understanding this app currently is working so that the user needs to select projects before getting recommendations. And instead of checking on the home page if the projects were chosen already and if not redirecting to project selection page, I decided to have project selection on the home page and then redirect to the page with recommendations. Do you think that is a bad design? Why?

frontend/src/router/index.ts
s17k: I guess the name Home is slightly misleading, we don't have a typical Home that acts as a navigation page (links to sub-pages you can open).
If we added graphs, we would put a link to them maybe in the upper app bar or somewhere near the table. But, we would still put recommendations table in '/'. If you go to [bbc.co.uk](http://www.bbc.co.uk), you don't just get links to "Sports", "World news" and "Entertainment". You also get some actual news :)

frontend/src/router/index.ts
s17k: Totally agree, we want the user to select projects before getting recommendations.
But, there is a reason that BBC doesn't have sign-in at /, they just put the news there as that is the actual core of their app.

frontend/src/views/Recommendations.vue
```diff
@@ -0,0 +1,52 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. -->
+
+<template>
+  <v-app>
+    <v-app-bar app color="primary" dark>
```
s17k, line 17: It would be nice not to have this code duplicated across views. What if we decide to change the colour or add a button there ?

Podkamienna: Yes but that is probably something for a different PR.

frontend/src/router/index.ts
Podkamienna: But we don't have sign in at /? Also they don't need sign in to display the news. And they don't need selecting cities to display only news for those cities.

frontend/src/views/Recommendations.vue
s17k: I disagree, you would have added less code by making this into a component :)

frontend/src/router/index.ts
s17k: Okay, let me try differently:

There is a benefit to designing the home page so that we redirect if we don't have something and then come back to it. This way you don't enforce a strict order of views and it makes it much easier to focus on what do we actually do, as the Home handler will look something like this:
- no token, get a token and come back
- no project selected, ask the user to select some projects, come back
... (maybe other stuff at some point)
So, instead of a strict order of views that is hard to follow (you would need to look at next() calls all over Home.vue), you would have a list of ifs.

frontend/src/views/Recommendations.vue
Podkamienna: Yes, if I thought about it before. But I came out with the idea only when I wanted to make "Recomator" a button, so I would be deleting the code. Also I think, that this could be just a nice, small PR and not confuse everyone by adding something more here.

s17k: that is fine, maybe add an issue then

frontend/src/router/index.ts
Podkamienna: But I think that the strict order of the views makes it easier to understand.

frontend/src/components/CoreTable.vue
```diff
@@ -160,7 +160,10 @@ export default class CoreTable extends Vue {
 
   // Checks if everything is selected on the current page
   areAllSelected(): boolean {
-    return this.selectableRows.every(item => this.isSelected(item));
+    return (
+      this.selectableRows.every(item => this.isSelected(item)) &&
+      this.selectableRows.length !== 0
```
minorczyka, line 165: Shouldn't you first check if `length != 0`?

frontend/src/components/ProjectsWithRequirements.vue
minorczyka: You can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment to merge `headers` with this.

frontend/src/router/index.ts
```diff
@@ -33,9 +35,19 @@ const routes: Array<RouteConfig> = [
         next({ name: "GoogleSignIn" });
         return;
       }
+
+      const projectString = window.localStorage.getItem("project_list");
```
minorczyka, line 39: It would be nice to extract it to separate dedicated function or at lest not use literal string "project_list" but some constant.

minorczyka (changes requested):

frontend/src/components/CoreTable.vue
s17k: It doesn't matter, every returns true for empty arrays: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

frontend/src/router/index.ts
```diff
@@ -99,11 +99,21 @@ const routes: Array<RouteConfig> = [
     path: "/homeWithInit",
     name: "HomeWithInit",
     beforeEnter(_, __, next) {
+      const projectString = window.localStorage.getItem("project_list");
+      if (projectString === null) {
```
s17k, line 103: What if it is undefined?

s17k: That is actually way more probable. localStorage getItem returns undefined if there is no key with such value

frontend/src/store/recommendations_store.ts
```diff
@@ -107,6 +107,7 @@ const actions: ActionTree<IRecommendationsStoreState, IRootStoreState> = {
     context.commit("resetRecommendations");
     context.commit("setProgress", 0);
 
+    console.log(context.rootGetters["selectedProjects"]);
```
s17k, line 110: You would have probably found it on your own, but let's make sure :)

frontend/src/views/Home.vue
```diff
@@ -50,5 +59,10 @@ import ProgressWithHeader from "@/components/ProgressWithHeader.vue";
 @Component({
   components: { CoreTable, Footer, ProgressWithHeader }
 })
-export default class Home extends Vue {}
+export default class Home extends Vue {
+  getProjectSelection() {
+    this.$store.commit("projectsStore/setSelected", [], {root: true});
```
s17k, line 64: What is `{root: true}` doing ?

frontend/src/router/index.ts
Podkamienna: I'm not sure, when I check in my browser's console it returns null.

Podkamienna: https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem

s17k: You are right! My bad.

frontend/src/views/Home.vue
Podkamienna: I guess that it is not necessary here, but what it does is calling a mutation as if the store it is called on was the root of the store. So you can call something from projectsStore, when having recommendationsStore.

s17k: yeah, you don't need it here then. I didn't know that there is such an option, cool!

s17k:
It would be amazing if you could post some screenshots, as you are adding a lot of new UI.
I will run this on my own as well, but it would be easier to discuss if everyone had it easily visible.

frontend/src/components/ProjectsWithRequirements.vue
```diff
@@ -0,0 +1,187 @@
...
+      })
+    )
+    .map(elt => {
+      if (elt.value === "name") {
+        return elt;
+      } else {
+        return Object.assign(elt, { align: "center" });
+      }
+    });
+
```
s17k, line 173: I have some ideas how to simplify this:
```typescript
headers = [
   {value:"name"}, 
    this.requirementList..map(reqName => {value: reqName, align: "center" })
];
   ```

File: frontend/tests/unit/home_view.spec.ts [hunk 18]
```diff
@@ -15,7 +15,7 @@ limitations under the License. */
 import Vue from "vue";
 import { mount } from "@vue/test-utils";
 import { rootStoreFactory } from "@/store/root_store";
-import Home from "@/views/Home.vue";
+import Recommendation from "@/views/Home.vue";
```
s17k, line 18: I think this should stay as 'Home' :)

s17k (changes requested):

Podkamienna:
image

Podkamienna:
image

Podkamienna:
image

Podkamienna:
image

s17k:
There is no nice way to stop fetching once we start it. So, imo the settings button should not be there when the recommendations are still being fetched.

s17k:
I am writing this app and it took me a while to figure out what "recommendations" and "requirements" buttons actually do. I would vote for naming them: 
- "See permissions" or "Permissions table" 
- "Confirm" or "OK" or "Finish"

s17k:
Also, let's maybe disable the requirements/recommendations buttons if there are no selected projects, to make this more intuitive ?

Podkamienna:
But if there was no setting button, then if someone added a new project and wanted to check if there is some recommendation for it, they would have to wait for all the recommendations to be fetched, which could take quite a long time. It actually didn't occur to me, that it could not work. But you are right and the simplest way to fix it is not allowing users to change the project while loading. I don't really like it though.

s17k:
They can just reload the whole page if they like :), so it's not like they absolutely need to wait once they choose their projects.

Podkamienna:
If they reload the page, they will have to wait anyway, as the selected projects are saved in the localstorage. What they could do is clearing the localstorage.

Podkamienna:
Updated photos of UI:
image
image
image
image

s17k:
Looks good. 
I would only vote for putting the black buttons together on the right, instead of having them spread out and centered.

frontend/src/components/ProjectsWithRequirements.vue
```diff
@@ -156,20 +156,12 @@ export default class ProjectList extends Vue {
...
-      })
-    )
-    .map(elt => {
-      if (elt.value === "name") {
-        return elt;
-      } else {
-        return Object.assign(elt, { align: "center" });
-      }
-    });
+  headers = ([] as { value: string; align?: string }[]).concat(
```
s17k, line 159: I think we can just use "as" on the first element `[{value: "name"}]` and avoid using concat.

frontend/src/components/AppBar.vue
```diff
@@ -0,0 +1,32 @@
...
+      <h1>Recomator</h1>
+    </v-btn>
+    <v-spacer />
+    <slot />
+  </v-app-bar>
+</template>
+
+<script lang="ts">
+import Vue from "vue";
+
```
s17k, line 26: Missing @Component

s17k, line 26: Missing `@Component` results in an error when I tried running this code on my own.

s17k (changes requested):
Please fix :)

frontend/src/components/AppBar.vue
Podkamienna: I'm sorry, I managed not to push my changes somehow, it should work now.

frontend/src/store/requirements_store.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      if (responseCode !== HTTP_OK_CODE) {
+        context.commit("setError", {
+          errorCode: responseCode,
+          errorMessage: `progress check failed: ${response.statusText}`
+        });
+
+        context.commit("endFetch");
+        return;
+      }
```
minorczyka, line 120: We should handle errors similarly to how it's done in Stan's new PR (https://github.com/googleinterns/recomator/pull/203). But it can stay as it is for now, let's do it in the later PR.

minorczyka (approved):

s17k (approved):

PR #203, opened 2020-09-21
New error page, authFetch handles failed requests

s17k:
There is a new error page, that looks like this:
Screenshot 2020-09-22 at 9 57 38 AM

- A short demo of what happens now when the backend doesn't exist, or is not reachable (quite common during set-up): https://screencast.googleplex.com/cast/NTQyNzgzNDEwMzI2NzMyOHxmZDhjZmI4NS1lYQ
What the user used see in this case was a "localhost:8079/redirect" as the page source (or whatever the hardcoded backend address was) in a failed tab in the browser.

- I also made a test: I turned off the wifi in the middle of fetching recommendations for about 2s. The fetch resumed without errors and showed a list of recommendations. That is new, we used to just kill the app without any notice to the user.

Closes: #188 
Closes: #197

### DIFF
--- a/frontend/src/components/ErrorMsg.vue
+++ b/frontend/src/components/ErrorMsg.vue
@@ -1,0 +1,80 @@
+<!-- Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. -->
+<template>
+  <v-app>
+    <v-app-bar app color="primary" dark>
+      <h1>Recomator</h1>
+    </v-app-bar>
+    <v-main>
+      <v-card id="errorCard" persistent max-width="700">
+        <v-card-title id="elem" class="justify-center">
+          <v-icon large color="error">mdi-alert-circle</v-icon>
+          Error: {{ header }}
+        </v-card-title>
+        <v-expansion-panels accordion multiple v-model="openedIndices">
+          <v-expansion-panel v-for="(value, name) in body" :key="name">
+            <v-expansion-panel-header>{{ name }}</v-expansion-panel-header>
+            <v-expansion-panel-content>
+              {{ value }}
+            </v-expansion-panel-content>
+          </v-expansion-panel>
+        </v-expansion-panels>
+      </v-card>
+    </v-main>
+  </v-app>
+</template>
+
+<script lang="ts">
+import Vue, { PropType } from "vue";
+import { Component } from "vue-property-decorator";
+
+const ErrorMsgProps = Vue.extend({
+  props: {
+    header: {
+      type: String,
+      required: true
+    },
+    body: {
+      type: Object as PropType<Record<string, string>>,
+      required: true
+    }
+  }
+});
+
+@Component
+export default class ErrorMsg extends ErrorMsgProps {
+  openedIndices: number[] = [];
+  mounted() {
+    const numberOfSections = Object.keys(this.body).length;
+
+    // [0,1,...,numberOfSections-1]
+    this.openedIndices = Array.from(Array(numberOfSections).keys());
+  }
+}
+</script>
+
+<style lang="scss">
+#errorCard {
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+}
+#elem {
+  -webkit-hyphens: none;
+  -moz-hyphens: none;
+  -ms-hyphens: none;
+  hyphens: none;
+}
+</style>
--- a/frontend/src/router/misc.ts
+++ b/frontend/src/router/misc.ts
@@ -23,3 +23,8 @@ export async function isBackendResponsive(): Promise<boolean> {
   }
   return true;
 }
+
+// Remember last selected project list in localStorage
+export function readProjectList() {
+  return window.localStorage.getItem("project_list");
+}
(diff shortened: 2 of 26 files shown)


frontend/src/components/ErrorMsg.vue
```diff
@@ -0,0 +1,80 @@
...
+
+@Component
+export default class ErrorMsg extends ErrorMsgProps {
+  openedIndices: number[] = [];
+  mounted() {
+    const numberOfSections = Object.keys(this.body).length;
+
+    // [0,1,...,numberOfSections-1]
+    this.openedIndices = Array.from(Array(numberOfSections).keys());
+  }
```
minorczyka, line 63: It's better to make `openedIndices` just a computed property (getter).

frontend/src/router/misc.ts
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */
+
+import { getBackendAddress } from "@/config";
+
+export async function isBackendResponsive(): Promise<boolean> {
+  // Let's first use the /auth endpoint to make sure that the backend exists
+  try {
+    await fetch(`${getBackendAddress()}/auth?code=123`);
```
minorczyka, line 20: Maybe we could call `/redirect` endpoint. So we don't make any unnecessary calls to Google APIs.

frontend/src/router/show_error.ts
```diff
@@ -0,0 +1,37 @@
...
+): Promise<void> {
+  if (!criticalError) {
+    console.log(["Warning:", header, body]);
+    return;
+  }
+
+  const headerParam = header ? header : "No name";
+  const bodyParam = body ? JSON.stringify(body) : JSON.stringify({});
+
+  window.location.assign(`/errorMsg?header=${headerParam}&body=${bodyParam}`);
```
minorczyka, line 32: We should encode this parameters (and later in router decode), we can't just pass these strings directly (https://en.wikipedia.org/wiki/Query_string).

frontend/src/store/auth/auth_fetch.ts
```diff
@@ -1,30 +1,84 @@
...
+    const initWithAuth = addAuth(init, rootState);
+
+    let response: Response;
+    let requestsLeft = maxRetries + 1;
+    while (requestsLeft > 0) {
+      requestsLeft--;
+      // wait fist if second or later attempt
+      if (requestsLeft < maxRetries) {
+        // typically gives us 3*2s for temporary internet problems
+        await delay(2000);
```
minorczyka, line 39: It's a good practice to use exponential backoff for retires (https://en.wikipedia.org/wiki/Exponential_backoff), e.g. for the first time we wait 2s, for the second time 4s, for the third time 8s etc.

If a service is having issues caused by a large amount of traffic, adding retires to clients makes things even worse. It works as a positive feedback which can potentially completely kill the service.

frontend/src/store/auth/auth_fetch.ts
```diff
@@ -1,30 +1,84 @@
...
+      } catch (error) {
+        // server unreachable or another serious problem with connection
+        const isCriticalError = requestsLeft === 0;
+        await showError(
+          `Failed to reach the Recomator backend`,
+          {
+            Message: error.Message,
+            URL: input,
+            Init: JSON.stringify(init),
+            "Stack trace": error.stack
```
minorczyka, line 53: In general you don't want to show stack trace to users, only log it somewhere. But if your audience is programmers then it might make some sense (some of our internal tools does it), it's also easier to report bugs.

I'm just not sure how the stack trace will look for minified JS code. You will need source maps to make any sense from them.

frontend/src/store/auth/auth_fetch.ts
```diff
@@ -1,30 +1,84 @@
...
+      // not successful (outside of 200-299) but server is responsive
+
+      const responseCode = response!.status;
+      // access denied
+      if (responseCode === 401 || responseCode === 403) {
+        // redirect to GoogleSignIn: this will shut down everyting,
+        // including fetching recommendations and check status
+        await rootState.router!.push({ name: "GoogleSignIn" });
+        // make sure we don't ever return from here
+        await delay(infiniteDurationMs);
```
minorczyka, line 71: That's a little dirty hack :smile:

minorczyka (changes requested):

frontend/src/store/auth/auth_fetch.ts
s17k: I was thinking something like: the user gets an error, makes an issue (ticket) and attaches the screenshot (with stack trace). So you don't even need to be a developer to make use of it.

frontend/src/router/misc.ts
s17k: That was my first idea, but once you do that it actually redirects your requests to a google sign-in page which makes some failed requests.

It does actually make Google API calls, I didn't think about that. Nice catch.
I will just change it to `/recommendations` which responds with `Authorization header not in the form 'Bearer...` so the backend will probably not be doing any Google API calls (which also makes sense, I guess it just checks if the token is present in memory).

frontend/src/store/auth/auth_fetch.ts
s17k: Yep, fair point.

frontend/src/router/misc.ts
minorczyka: I thought it won't make redirection if we call it just with fetch, I'm surprised. But `/recommendaitons` without auth headers also sounds good.

frontend/src/store/auth/auth_fetch.ts
minorczyka: Sounds good, it can stay.

frontend/src/router/show_error.ts
s17k: Yes, I was really worried about that. But it magically got encoded (I saw the browser address) and decoded without anything explicit so I gave in.

frontend/src/components/ErrorMsg.vue
s17k: It's a two-way binding though.

minorczyka: Ok, I haven't noticed.

minorczyka (approved):

PR #206, opened 2020-09-24
Added styling to avoid line breaking

Podkamienna:
Before:
Screenshot 2020-09-23 at 11 56 25 AM
After:
Screenshot 2020-09-23 at 2 05 03 PM
 I wasn't able to get breaking lines in the text of the error. Moreover, both errors that we have in the recommendations right now are empty.


s17k (approved):
Nice, I was looking for something like that for the error message component as well.

PR #211, opened 2020-09-24
Speedup

ViktoriiaKovalova:


pkg/automation/access.go
```diff
@@ -40,29 +41,30 @@ type Requirement struct {
...
+						ErrorMessage: googleErr.Message,
+					}}, nil
+				}
+			} else {
+				result = []*Requirement{&Requirement{
+					Name:      serviceUsageName,
+					Satisfied: true,
+				}}
+			}
+		}
```
minorczyka, line 67: Can you add comment here about what we're doing here? That we skip checking Service Usage API availability and just deduce it from first request send to it.

minorczyka (approved):

PR #213, opened 2020-09-24
Fix: project list can land at an invalid page #209

s17k:
Project list used to keep the original page number, even if it didn't exist after applying the search filter. 
Now, whenever we change the search text, the page number is set to 1.

Demo (after): https://screencast.googleplex.com/cast/NTc1NTY2MzExODQ5OTg0MHw4ZTU2OTQ1NC1lMw

Fixes: #209


frontend/src/components/ProjectList.vue
```diff
@@ -124,6 +126,8 @@ export default class ProjectList extends Vue {
   searchEnabled = false;
   search = "";
 
+  page = 1;
```
Podkamienna, line 129: You could use resetPage here probably, but the rest is great.

Podkamienna (approved):

frontend/src/components/ProjectList.vue
s17k: This is a declaration.

Podkamienna: Ymm, sorry, you are right.

PR #225, opened 2020-10-01
fix frontend freezing while listing projects

ViktoriiaKovalova:


frontend/src/store/projects_store.ts
```diff
@@ -25,12 +25,12 @@ import {
 const BACKEND_ADDRESS: string = getBackendAddress();
 
 const mutations: MutationTree<IProjectsStoreState> = {
-  addProject(state, project: string): void {
-    if (state.projects.filter(elt => elt.name === project).length !== 0) {
-      throw "Duplicate recommendation name";
+  setProjects(state, projects: string[]): void {
+    if (new Set(projects).size !== projects.length)
+      throw "Duplicates found among given projects' names";
```
minorczyka, line 30: I think we shouldn't throw error when we find duplicates, it's fine to just ignore them, maybe only log some warning.

minorczyka (approved):

frontend/src/store/projects_store.ts
ViktoriiaKovalova: is it fine to leave console.log()? cause I think there are some warnings to remove it, but it's nice to know what happened

PR #226, opened 2020-10-01
add cloud build

ViktoriiaKovalova:


.gcloudignore
```diff
@@ -23,5 +23,3 @@
 *.test
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
-
-frontend/
```
minorczyka, line 27: Uploading `frontend` directory is unnecessary, especially I would like to make sure that we won't upload `node_modules` directory because it's very heavy.

deploy_gcloudignore.txt
```diff
@@ -0,0 +1,2 @@
+frontend/*
+!frontend/dist
```
minorczyka, line 2: What's the reason for having it in separate file?

minorczyka (changes requested):

deploy_gcloudignore.txt
ViktoriiaKovalova: first, l it's used by cloud build (it creates config.json and .env file with backend address from Secret Manager API, then builds frontend). at this point we need /frontend folder.
then, cloud build goes to deployment step(running gcloud app deploy) and we don't need frontend/ folder, so we need to update .gcloudignore
So, the point is .gcloudignore is used 2 times and at the first time we need frontend folder, at the second time - not. So .gcloudignore is updated after frontend is built.

ViktoriiaKovalova: You can see lines 17-25 of cloudbuild.yaml  - it's the step which updates .gcloudignore

.gcloudignore
ViktoriiaKovalova: we don't upload it to app engine, it's the .glcoudignore which is used when cloning repository for build

deploy_gcloudignore.txt
minorczyka: When Cloud Build clones repository it removes files that are present in `.cloudignore`?

minorczyka (approved):

PR #234, opened 2020-10-06
Retry quota exceeded

ViktoriiaKovalova:


pkg/automation/disks.go
```diff
@@ -73,8 +77,12 @@ func (s *googleService) CreateSnapshot(project, zone, disk, name string) error {
...
-	err := AwaitCompletion(func() (*compute.Operation, error) {
-		return disksService.Delete(project, zone, disk).RequestId(requestID).Do()
-	}, sleepTimeDeletingDisks)
+	var err error
+	DoRequestWithRetries(func() error {
+		err = AwaitCompletion(func() (*compute.Operation, error) {
+			return disksService.Delete(project, zone, disk).RequestId(requestID).Do()
+		}, sleepTimeDeletingDisks)
+		return err
+	})
```
minorczyka, line 86: You can make `DoRequestWithRetries` return an error, it will make writing this code a little bit more cleaner, you won't have to define `var err` and will be able to return err without assignment in the lambda.

minorczyka (changes requested):

pkg/automation/disks.go
ViktoriiaKovalova: I don't know how to do it:
```
var response *serviceusage.GoogleApiServiceusageV1Service
		err := DoRequestWithRetries(func() error {
			response, err := servicesService.Get("projects/" + project + "/services/" + api).Do()
			return err
		})
```
this way i get:
```
../../pkg/automation/access.go:52:4: response declared but not used
```
And this way:
```
var response *serviceusage.GoogleApiServiceusageV1Service
		err := DoRequestWithRetries(func() error {
			response, err = servicesService.Get("projects/" + project + "/services/" + api).Do()
			return err
		})
```
I get
```
../../pkg/automation/access.go:52:14: undefined: err
../../pkg/automation/access.go:53:11: undefined: err
```

minorczyka: In this case it will really be change for worse. You will have to do it like:
```
var response *serviceusage.GoogleApiServiceusageV1Service
err := DoRequestWithRetries(func() error {
	resp, err := servicesService.Get("projects/" + project + "/services/" + api).Do()
	response = resp
	return err
})
```

ViktoriiaKovalova:
@minorczyka there's a test checking that after receiving some error codes(302, 405, 503) frontend will retry, what should I do with it?

minorczyka (approved):

PR #236, opened 2020-10-06
add redirect to google sign in 

ViktoriiaKovalova:


frontend/src/router/index.ts
```diff
@@ -17,41 +17,33 @@ import VueRouter, { RouteConfig } from "vue-router";
...
+  return true;
+}
+
 const routes: Array<RouteConfig> = [
   {
     path: "/",
     name: "Home",
     component: Home,
     async beforeEnter(_, __, next): Promise<void> {
-      if (!(await isBackendResponsive())) {
```
minorczyka, line 35: What error we will exactly get when backend is down? Because it was intentionally added in https://github.com/googleinterns/recomator/pull/203. Maybe we don't really need this, I just don't remember how it behaved previously.

ViktoriiaKovalova: probably it's done because it's redirected to backend /redirect and in that case we'll see some browser error if backend is not responding. so i'll return it and put it in /googleSignIn

minorczyka (approved):

PR #239, opened 2020-10-09
Internal zones fix

ViktoriiaKovalova:


pkg/automation/recommendations.go
```diff
@@ -43,12 +49,27 @@ func (s *googleService) ListRecommendations(project, location, recommenderID str
...
-	return recommendations, nil
+	err := DoRequestWithRetries(func() error {
+		recommendations = nil
+		err := listCall.Pages(s.ctx, addRecommendations)
+		// Check if error is because current location is not available for getting recommendations.
+		if googleErr, ok := err.(*googleapi.Error); ok {
+			body := googleErr.Body
+			var fields map[string]interface{}
+			if parseErr := json.Unmarshal([]byte(body), &fields); parseErr != nil {
+				return err
```
minorczyka, line 60: Let's not return this error, it's better to return the original error.

pkg/automation/recommendations.go
```diff
@@ -43,12 +49,27 @@ func (s *googleService) ListRecommendations(project, location, recommenderID str
...
+			}
+			if errorJSON, ok := fields["error"].(map[string]interface{}); ok {
+				if status, ok := errorJSON["status"].(string); ok && status == "INVALID_ARGUMENT" {
+					log.Printf("Invalid location error: %v received while getting recommendations for %s %s %s",
+						body, project, location, recommenderID)
+					return nil
+				}
+			}
```
minorczyka, line 68: Can we move this check to separate function? It would make `ListRecommendations` function more readable.

minorczyka (changes requested):
I can see changes from the second PR.

pkg/automation/recommendations.go
ViktoriiaKovalova: it's the original error. I made a check with return here to avoid having many nested ifs.

minorczyka (approved):

